Issue #17, opened 2020-12-04
enumopencl sample program crashes


kenba:
I built the SDK on Windows 10 using cmake 3.19.1, MSVC 2019 and OpenCL.lib from AMD APP SDK 3.0.
When I ran the enumopencl sample program, MSVC raised the following error:

![image](https://user-images.githubusercontent.com/3299660/101153918-ec842700-361c-11eb-84bd-c8aecd3d2916.png)

The debugger showed the program to be failing on line 111 in `AllocateAndGetDeviceInfoString`: `delete[] param_value;`  
Tracing back to `PrintDeviceInfoSummary`, it seems the crash was caused by `deviceProfile = NULL;` being missing after `delete[] deviceProfile;` causing line 111 in `AllocateAndGetDeviceInfoString` to delete the deviceProfile array twice.

PR #18, opened 2020-12-04
Fix enumopencl for issue #17

kenba:
Fix issue #17 by setting `deviceProfile = NULL;` in the code block with the other char arrays.


bashbaug (approved):
Thanks for catching this!  We added the device profile at the last minute and obviously missed this part of it.

kenba:
My pleasure. I’ve been using C++ and Rust for so long, I forgot how difficult memory management is in C!

PR #21, opened 2021-01-13
De-duplicate CMake logic for samples

kpet:
Signed-off-by: Kevin Petit <kevin.petit@arm.com>

### DIFF
--- a/samples/CMakeLists.txt
+++ b/samples/CMakeLists.txt
@@ -12,4 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+if(CMAKE_CONFIGURATION_TYPES)
+    set(SAMPLE_CONFIGS ${CMAKE_CONFIGURATION_TYPES})
+else()
+    set(SAMPLE_CONFIGS ${CMAKE_BUILD_TYPE})
+endif()
+
+# Usage: add_sample(name source0 [source1] ...)
+function(add_sample name)
+  add_executable(${name} ${ARGN})
+  target_include_directories(${name} PRIVATE ${OPENCL_SDK_INCLUDE_DIRS})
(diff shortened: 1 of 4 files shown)


MathiasMagnus:
The current control of install layout seems more code than I believe it could be (my guts are telling me). It is tangential to the deduplication at hand, so I'll defer "fixing" that once I come with a better way. (Multi-config generators should be taking care of this automatically, or generator expressions to install() should do the trick. No need for if()s and foreach()s. (At least that's the theory)

MathiasMagnus (approved):

kpet:
Thanks for the review! Yes, agree this can probably be made simpler. I was just looking to reduce the duplication first.

bashbaug (approved):
This is a nice improvement and can be merged anytime.

FWIW, I went through a similar de-duplication recently in some of my personal samples.  A few enhancements I've found useful that we may want to make shortly are:

1. Move the OpenCL version setting into the CMake function as well to reduce boilerplate.
2. Add the ability to install additional files, such as files containing OpenCL kernels.
3. Add the ability to link with additional libraries, say for OpenGL visualization or utility functions.

In case it's helpful, see:

* [CMake Function](https://github.com/bashbaug/SimpleOpenCLSamples/blob/master/samples/CMakeLists.txt#L29)
* [Example Usage](https://github.com/bashbaug/SimpleOpenCLSamples/blob/master/samples/00_enumopencl/CMakeLists.txt#L21)

kpet:
+1 on @bashbaug's suggestions. That's the way to go!

kpet:
I've created https://github.com/KhronosGroup/OpenCL-SDK/issues/22 to cover enhancing the function further as per Ben's suggestions. Merging this PR as is.

PR #32, opened 2021-05-13
Add OpenCL ruby samples

Kerilk:
This pull request adds Ruby samples reproducing the C/C++ samples already present in the SDK.
It also adds CI on Ubuntu and MacOS. If an OpenCL platform could be installed on windows, we could test it there as the bindings have been reported to work on windows as well.

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -91,8 +91,8 @@ jobs:
             time python $i
           done
 
-  buildruby:
-    name: Build Ruby ${{ matrix.os }}
+  checkruby:
+    name: Check Ruby Samples ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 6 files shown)


bashbaug (approved):
This is great!  I was able to execute all of these samples and they behave as expected.

My one suggestion is to add a README describing what is needed to run these samples, or a link to an appropriate Ruby document.  I was able to figure it out from the CI steps (I didn't have `opencl_ruby_ffi` on this sytem) but it would be nicer to have it in a narrative document.

Thank you for your contribution!

Kerilk:
@bashbaug Thanks for the review. I added a (simple) README.md describing the scope of the samples and instructions to install `opencl_ruby_ffi` bindings. We can expand it if we deem it necessary.

bashbaug:
There's a GitHub actions conflict in this PR now.  Looks like it should be easy to resolve.  @Kerilk I'll do that this afternoon, unless you get to it first - thanks!

Kerilk:
@bashbaug Thanks for the heads up, should be fine now.

PR #33, opened 2021-06-30
bump external submodules

bashbaug:


alycm (approved):
Passes CI, I also built and ran the samples locally, I confirmed that the commits are to the latest ones.

Will merge immediately.

PR #37, opened 2021-12-03
Samples and Utility libraries

MathiasMagnus:
This PR ships a larger batch of samples and the initial set of C/C++ utility libraries and SDK libraries. This PR includes:

- Temporary redirection of git submodules to the related Stream HPC SDK repos. This PR is dependent on changes found in:
  - [OpenCL-SDK base changes in OpenCL-Headers](https://github.com/KhronosGroup/OpenCL-Headers/pull/189)
  - [OpenCL-SDK base changes in OpenCL-ICD-Loader](https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/160)
  - [OpenCL-SDK base changes in OpenCL-CLHPP](https://github.com/KhronosGroup/OpenCL-CLHPP/pull/168)
- References to existing and new chapters of the OpenCL-Guide:
  - [Getting Started Guide: Windows & CMake support](https://github.com/KhronosGroup/OpenCL-Guide/pull/16)
  - [Initial OpenCL-OpenGL interop overview](https://github.com/KhronosGroup/OpenCL-Guide/pull/15)
- The batch of samples include:
  - SAXPY (C & C++)
  - Reduction (C & C++)
  - Histogram _soon_ (C++)
  - Blur (C)
  - Load/Save Binaries (C)
  - Conway's Game of Life (C++)
  - NBody (C++)
- All samples are accompanied by README markdown detailing the purpose of the samples, their feature set, some of their internal working and the utilized API surface.
- Two libraries are built within the SDK
  - OpenCLUtils(Cpp) which consumers of the SDK can use to facilitate some common OpenCL tasks
    - The utility libraries only depend on other OpenCL packages
  - OpenCLSDK(Cpp) which the samples use to deduplicate some of the common aspects of SDK samples, such as command-line parsing and window handling
    - The OpenCLSDK libraries have more dependencies, but nothing out of ordinary
    - The windowing library of choice is SFML, because it has explicit support for Android, and that is a platform the SDK wishes to target along with the native samples
- Building the development packages only don't require any dependencies, a non-recursive clone with submodules is enough. This is tested in CI.

Some notes on the shape of the SDK samples and some of the design choices:

- The C dependent libraries are bundled within the source tree. This was done because while Linux typically has these libraries available via the system package managers, users of Windows will have to rely on C/C++ package managers like Conan/Vcpkg/Hunter/etc. Once the availability of the deps improve, removing them from the repository may be appropriate.
  - Although it is a contentious topic, using Git submodules as a form of package management isn't ideal, nor is bundling deps in the source tree. While pkg managers gain popularity, best case scenario is catering to both crowds: those not using pkg managers but wanting a clean build experience and those who readily have custom built versions of all their deps, often newer than those bundled.
  - Being an author of various Vcpkg ports myself and maintaining HIP libraries, I know for a fact that self-hosting and wanting to provide a clean build experience for 3 platforms has *non-zero* cost. Offloading that pain to package manager maintainers is the best and easiest approach.
    - Do note that once the CMake rework of the SDK goes through, one of my first tasks will be to simplify the [OpenCL Vcpkg port](https://github.com/microsoft/vcpkg/blob/master/ports/opencl/portfile.cmake).
   - TCLAP has not been bundled. Should it be decided that the SDK should bundle that as well, we can add it next to the other deps.


README.md
```diff
@@ -13,40 +13,46 @@ It also contains resources useful to OpenCL developers:
...
+       git submodule init
+       git submodule update
+
+1. Install dependencies
+
+       vcpkg install sfml tclap glm
+
+1. Build and install SDK with samples and no downstream unit tests
+
+       cmake -D BUILD_TESTING=OFF -D BUILD_DOCS=OFF -D BUILD_EXAMPLES=OFF -D BUILD_TESTS=OFF -D OPENCL_SDK_BUILD_SAMPLES=ON -D OPENCL_SDK_TEST_SAMPLES=OFF -D CMAKE_TOOLCHAIN_FILE=/vcpkg/install/root/scripts/buildsystems/vcpkg.cmake -B ./OpenCL-SDK/build -S ./OpenCL-SDK
```
bashbaug, line 57: Suggest breaking up this long line to ease readability:

```suggestion
       cmake -D BUILD_TESTING=OFF \
               -D BUILD_DOCS=OFF \
               -D BUILD_EXAMPLES=OFF \
               -D BUILD_TESTS=OFF \
               -D OPENCL_SDK_BUILD_SAMPLES=ON \
               -D OPENCL_SDK_TEST_SAMPLES=OFF \
               -D CMAKE_TOOLCHAIN_FILE=/vcpkg/install/root/scripts/buildsystems/vcpkg.cmake \
               -B ./OpenCL-SDK/build -S ./OpenCL-SDK
```

It might also be a good idea to use a generic path like `/path/to/your/vcpkg.cmake` since this will change depending on how vcpkg is installed.

lib/src/Utils/File.c
```diff
@@ -0,0 +1,271 @@
...
+        if (err == CL_INVALID_VALUE) {
+            fprintf(stderr, "No suitable file %s found\n", filename);
+            goto end;
+        }
+        else if (err != CL_SUCCESS)
+            goto end;
+    }
+
+    OCLERROR_PAR(program = clCreateProgramWithBinary(context, num_devices, devices,
+        binaries_size, binaries_ptr, &err, NULL), err, end);
```
bashbaug, line 259: Heads up, I'm getting a warning here:

```
In file included from /home/bashbaug/git/OpenCL-SDK/lib/include/CL/Utils/Utils.h:6,
                 from /home/bashbaug/git/OpenCL-SDK/lib/src/Utils/Utils.c:2:
/home/bashbaug/git/OpenCL-SDK/lib/src/Utils/File.c: In function ‘cl_util_read_binaries’:
/home/bashbaug/git/OpenCL-SDK/lib/src/Utils/File.c:259:24: warning: passing argument 5 of ‘clCreateProgramWithBinary’ from incompatible pointer type [-Wincompatible-pointer-types]
  259 |         binaries_size, binaries_ptr, &err, NULL), err, end);
      |                        ^~~~~~~~~~~~
      |                        |
      |                        unsigned char **
/home/bashbaug/git/OpenCL-SDK/lib/include/CL/Utils/Error.h:35:6: note: in definition of macro ‘OCLERROR_PAR’
   35 | do { func; if (err != CL_SUCCESS) goto label; } while (0)
      |      ^~~~
In file included from /home/bashbaug/git/OpenCL-SDK/lib/include/CL/Utils/Error.h:13,
                 from /home/bashbaug/git/OpenCL-SDK/lib/include/CL/Utils/Utils.h:6,
                 from /home/bashbaug/git/OpenCL-SDK/lib/src/Utils/Utils.c:2:
/home/bashbaug/git/OpenCL-SDK/external/OpenCL-Headers/CL/cl.h:1244:58: note: expected ‘const unsigned char **’ but argument is of type ‘unsigned char **’
 1244 |                           const unsigned char **         binaries,
      |                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
```

samples/extensions/khr/conway/main.cpp
```diff
@@ -0,0 +1,366 @@
...
+    glBindBuffer(GL_ARRAY_BUFFER, vertex_buffer); checkError("glBindBuffer(GL_ARRAY_BUFFER, vertex_buffer)");
+        glBufferData(GL_ARRAY_BUFFER, quad.size() * sizeof(float), quad.data(), GL_STATIC_DRAW); checkError("glBufferData(GL_ARRAY_BUFFER, quad.size() * sizeof(float), quad.data(), GL_STATIC_DRAW)");
+    glBindBuffer(GL_ARRAY_BUFFER, 0); checkError("glBindBuffer(GL_ARRAY_BUFFER, 0)");
+
+    glGenVertexArrays(1, &vertex_array); checkError("glGenVertexArrays(1, &vertex_array)");
+    glBindVertexArray(vertex_array); checkError("glBindVertexArray(vertex_array)");
+        glBindBuffer(GL_ARRAY_BUFFER, vertex_buffer); checkError("glBindBuffer(GL_ARRAY_BUFFER, vertex_buffer)");
+        glVertexAttribPointer(0, 2, GL_FLOAT, GL_FALSE, 4*sizeof(float), (GLvoid *)(NULL)); checkError("glVertexAttribPointer(0, 2, GL_FLOAT, GL_FALSE, 4*sizeof(float), (GLvoid *)(NULL))");
+        glVertexAttribPointer(1, 2, GL_FLOAT, GL_FALSE, 4*sizeof(float), (GLvoid *)(NULL + 2 * sizeof(float))); checkError("glVertexAttribPointer(1, 2, GL_FLOAT, GL_FALSE, 4*sizeof(float), (GLvoid *)(NULL + 2 * sizeof(float)))");
```
bashbaug, line 198: I'm getting a warning here too:

```
/home/bashbaug/git/OpenCL-SDK/samples/extensions/khr/conway/main.cpp: In member function ‘virtual void Conway::initializeGL()’:
/home/bashbaug/git/OpenCL-SDK/samples/extensions/khr/conway/main.cpp:198:108: warning: NULL used in arithmetic [-Wpointer-arith]
  198 |         glVertexAttribPointer(1, 2, GL_FLOAT, GL_FALSE, 4*sizeof(float), (GLvoid *)(NULL + 2 * sizeof(float))); checkError("glVertexAttribPointer(1, 2, GL_FLOAT, GL_FALSE, 4*sizeof(float), (GLvoid *)(NULL + 2 * sizeof(float)))");
      |  
```

bashbaug (changes requested):
Thanks for pulling this all together!  This is a huge upgrade to the SDK.

I haven't done an in-depth review but I did do a quick skim through the documentation and identified a few issues.

I also managed to build the updated SDK including all samples following the attached build instructions on my Ubuntu install.  Getting vcpkg setup wasn't entirely trivial, but once it was done things pretty much "just worked".  I flagged the few build warnings I saw.

The biggest issue I have after building and installing is that a lot of samples don't run due to missing libraries (I've installed into my own install directory, underneath the SDK build tree).  Should we build the OpenCL utils as a static library or copy it into the same directory as the samples so this is more robust?

```
bashbaug@bashbaug-nuc:~/git/OpenCL-SDK/install/Release$ ./saxpy 
./saxpy: error while loading shared libraries: libOpenCLUtils.so: cannot open shared object file: No such file or directory
```

I can run the samples out of the build directory, where the executables still have the libs RPATH'd, but this has other problems.  Some examples:

saxpy fails with `CL_INVALID_VALUE`, though I don't see anywhere we are returning this error:

```
bashbaug@bashbaug-nuc:~/git/OpenCL-SDK/build/bin$ ./saxpy 
Selected platform by Intel(R) Corporation
Selected device: Intel(R) HD Graphics 620 [0x5916]
OpenCL C 3.0 


Error: CL_INVALID_VALUE
```

cppsaxpy can't find the kernel source, which is understandable since I am running from the build directory, but it isn't in the install directory either:

```
bashbaug@bashbaug-nuc:~/git/OpenCL-SDK/build/bin$ ./cppsaxpy 
Selected platform: Intel(R) Corporation
Selected device: Intel(R) HD Graphics 620 [0x5916]

Error: Cannot open kernel source: ./saxpy.cl
```

binaries cannot find the right binary, but I'm not sure how to create it:

```
bashbaug@bashbaug-nuc:~/git/OpenCL-SDK/build/bin$ ./binaries 
Selected platform by Intel(R) Corporation
Selected device: Intel(R) HD Graphics 620 [0x5916]
OpenCL C 3.0 

No suitable file Collatz-Intel(R) HD Graphics 620 [0x5916].bin found

Error: CL_INVALID_VALUE
```

How can we improve the odds that this all "just works" and/or improve diagnostics when an error occurs?

samples/extensions/khr/conway/main.cpp
MathiasMagnus: My colleague caught that one too. I'll change it to 0 to silence the warning. The OpenGL spec explicitly says that the 0 pointer is the beginning of the buffer.

> Specifies a offset of the first component of the first generic vertex attribute in the array in the data store of the buffer currently bound to the GL_ARRAY_BUFFER target. The initial value is 0.

Because it's a pointer, and the value is 0, I wanted to make it explicit by using `NULL` but the warning triggers.

.gitmodules
```diff
@@ -1,9 +1,12 @@
 [submodule "loader"]
 	path = external/OpenCL-ICD-Loader
-	url = https://github.com/KhronosGroup/OpenCL-ICD-Loader.git
+	url = git@projects.streamhpc.com:Khronos/OpenCL-ICD-Loader.git
```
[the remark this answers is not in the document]
MathiasMagnus: The paths currently point to https://github.com/StreamHPC/... because they all need CMake changes. As the "SDK base changes" PRs are merged, they can be reverted to the KhronosGroup hosted repos.

README.md
MathiasMagnus: Ah, thanks. I'm still at war with the single-line, multi-line YAML syntaxes of GitLab/GitHub. Sprinkle some Dockerfile syntax on top and confusion is maxed out. I just accepted my fate, but this is nicer.

MathiasMagnus:
@bashbaug I think I know where your woes are coming from (and both are totally legit).

1. Not finding the library happens only with multi-config generators. Building using the default Unix Makefiles or Ninja works. Working on a fix.
2. The headers is a sticky subject. (Semi-extensive exploration of the phase space is [here](https://github.com/MathiasMagnus/OpenCL-KernelLoad), which I intend on merging into either the OpenCL-Guide or the SDK.) The 'current working directory' (henceforth cwd)-relative solution is simple, but is fairly brittle. _Copying_ the kernels into the build tree is problematic because

  - One may copy at configuration time using `file(COPY ...)` or `configure_file()` but then one needs to reconfigure to update the files on disk. (It's annoying to develop samples that way.)
  - One may set it as a post-build step, but then it will only trigger when one modifies the _host_ code. (This is annoying too.)
  - One may promote it to a build step, in which case utility targets like `saxpy-kernel` and `nbody-kernel-shader` start mildly polluting the list of available targets. Samples run fine from the install tree, because install commands can properly track the timestamps of both the executables and the kernel files.

_Symlinking_ the kernels would be awesome, as it follows updates to code in the source tree, but that has many problems too:
  - The problem with that 🥁🥁🥁🥁🥁🥁🥁🥁 [`file(CREATE_LINK ...)`](https://cmake.org/cmake/help/latest/command/file.html#create-link) is CMake 3.14. (And `cmake -E create_symlink` as non-UNIX-only is 3.13.) We can raise minimum version when building the samples to 3.13 / 3.14 or accept the bloat
```cmake
if(CMAKE_SYSTEM_NAME MATHCES Windows)
  execute_process(COMMAND "mklink ...")
else()
  execute_process(COMMAND "ln -s ...")
endif()
```
**but even then**, symlinking on Windows [without admin rights requires developer mode](https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/) enabled on the system. 🤯 Wow...


Ultimately I accepted the fact that running the samples from the build tree requires making the cwd the path to the kernel in the source tree, but if it bugs anyone, pick an item from the above list or the samples should move away from the CWD-relative location of device code.

samples/extensions/khr/conway/main.cpp
MathiasMagnus: Fixed.

lib/src/Utils/File.c
MathiasMagnus: Fixed.

MathiasMagnus:
All samples now execute correctly from the build tree as well as the install tree in both single and multi-config generators.

bashbaug (commented):
Thanks, this is looking a lot better now!

I was able to run all of the samples save histogram on Linux, and I was able to run this one when I manually copied the kernel file to the install directory.  I wasn't able to test the CL-GL sharing samples because we don't support the interop extensions on Linux, but I was able to verify that they run and do not have any missing libraries.

I'm having trouble building on Windows though.  I think this is a problem on my end related to vcpkg (the OpenCL-SDK project still can't find TCLAP), but it hasn't been as smooth as Linux.  I'll try more tomorrow, since we do support the interop extensions on Windows.

bashbaug (approved):
I've been able to successfully build and run all apps on both  Windows and Linux - thanks @MathiasMagnus for all of your help getting this working!

For record keeping, the issues I was having on Windows were:

1. vcpkg was defaulting to the 32-bit x86 packages instead of the 64-bit x64 packages.  Fixed by explicitly specifying `--triplet x64-windows` for vpkg (see https://github.com/KhronosGroup/OpenCL-SDK/pull/37/commits/a64a4cbef33bd659ad9f3dad11f3e46bfda5006b).
2. After this, the samples built, but not all DLLs were being properly copied to the output directory.  Fixed by updating my cmake version (see https://github.com/KhronosGroup/OpenCL-SDK/pull/37/commits/9dd7c322c40903a7d40a050f9f8898fef1e06077).

samples/core/binaries/main.c
```diff
@@ -0,0 +1,317 @@
...
+    { // if binary not present, compile and save
+        const char *kernel_location = "./Collatz.cl";
+        char *kernel = NULL;
+        size_t program_size = 0;
+        char *options = NULL;
+
+        OCLERROR_PAR(kernel = cl_util_read_text_file(kernel_location,
+                                                     &program_size, &error),
+                     error, cont);
+        printf("OpenCL file red... ");
```
alycm, line 202:
```suggestion
        printf("OpenCL C file read... ");
```

.github/workflows/presubmit.yml
```diff
@@ -28,17 +28,25 @@ jobs:
           submodules: recursive
       - name: Build
         run: |
-            mkdir build
-            cd build
+            if [[ "${{ matrix.os }}" == "ubuntu-20.04" ]]; then
+              sudo apt update
+              sudo apt install -y libtclap-dev libglm-dev libglew-dev libsfml-dev libstb-dev libidn11 libx11-dev libxrandr-dev libxi-dev mesa-common-dev libgl1-mesa-dev libglu1-mesa-dev libudev-dev
```
alycm, line 33: I know Ubuntu isn't universal, but it might be helpful to put this in the readme.  Or at least the ones that aren't there by default on a desktop install.

I had to install these to get the SDK to build:

```sh
sudo apt install libtclap-dev libstb-dev libglew-dev libsfml-dev libglm-dev
```

.gitlab/stream-ci.yaml
```diff
@@ -0,0 +1,376 @@
...
+      -D CMAKE_TOOLCHAIN_FILE=/opt/Microsoft/vcpkg/scripts/buildsystems/vcpkg.cmake
+      -D CMAKE_C_COMPILER=$C_COMPILER
+      -D CMAKE_CXX_COMPILER=$CXX_COMPILER
+      -D CMAKE_BUILD_TYPE=Release
+      -D BUILD_TESTING=ON
+      -D ENABLE_OPENCL_LAYERS=ON
+      -D OPENCL_ICD_LOADER_BUILD_TESTING=ON
+      -D BUILD_DOCS=OFF
+      -D BUILD_EXAMPLES=OFF
+      -D BUILD_TESTS=ON
```
alycm, line 198: Should some of these be set in the CMakeLists.txt?

This line in particular I had to set to get anything to build (I suspect that is actually a bug in OpenCL-CLHPP CMake, but it wasn't immediately clear why -- I'll look again later).

Other variables are not essential, but setting them saves building unnecessary tests.

alycm (commented):
I haven't reviewed this PR.   I did build and test the samples on this branch, on an implementation with no GL interop running on an x86 CPU.

samples/core/binaries/main.c
MathiasMagnus: Added.

.github/workflows/presubmit.yml
MathiasMagnus: These packages are mentioned in the OpenCL-Guide in the Linux guide [here](https://github.com/KhronosGroup/OpenCL-Guide/pull/17/files#diff-89767f088390b2e71cfc4717bbfe10a989e2d42346d89e34e5b76510992845b7R124). These are only required if one builds the samples and not otherwise.

.gitlab/stream-ci.yaml
MathiasMagnus: Yes, I was thinking hard about making that also part of these PRs, but... scope creep. I really want to make the test option variables consistent. Currently only the ICD-Loader is able to selectively disable it's tests via [`OPENCL_ICD_LOADER_BUILD_TESTING`](https://github.com/KhronosGroup/OpenCL-ICD-Loader/blob/master/CMakeLists.txt#L136-L141) and that's something the CLHPP repo should enable as well. One may want to test the SDK samples, but not build the few hundred header tests. I'll see to it shortly after the SDK and the base changes go through.

.github/workflows/presubmit.yml
bashbaug: I'm guessing this is only needed if the package dependencies aren't satisfied via vcpkg?

Regardless, I agree it would be worth including this as an alternative in the README.  Installing via the OS package manager will likely be smoother than vcpkg (getting it working on Ubuntu wasn't trivial) and this is the mechanism we're testing in CI.

MathiasMagnus: One can install them either way. It's user preference.

PR #39, opened 2022-01-23
Ignore generated files from CMake, VS, XCode

LeonBrands:
This makes building the CMake project inside of the source folder (instead of inside of the build/ folder) more convenient

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,24 @@ install/
 # vim
 *~
 .*.sw[op]
+
+# MacOS
+.DS_Store
+*.xcodeproj/
+
+# Visual Studio
+*.vcxproj
+*.vcxproj.filters
+*.vcxproj.user
+*.sln
+
+# CMake
+CMakeFiles/
+CMakeScripts/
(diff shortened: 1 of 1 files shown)


MathiasMagnus:
Summoning @bashbaug and/or @kpet to share their thoughts, as the contribution in question is mildly contentious.

I'm personally reluctant to provide any support for doing in-source builds as they are universally discouraged, at least as far as CMake goes. If in-source builds break for someone and they point to us "giving consent", those aren't fixes I would like to handle.

Making the in-source builds less painful are totally doable on your end as well. Placing these into the cloned project under `.git/info/exclude` it will persist on your end without having to update `.gitignore` (This is what I do on my end when it's needed). If in-source builds are something you regularly do, I'd suggest registering a user `.gitignore` file as detailed [here](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/).

LeonBrands:
The maintenance point is very understandable 🙂 I'll close this and handle it locally instead.

PR #40, opened 2022-02-06
Auto-fetch missing dependencies (console samples)

MathiasMagnus:
This is basic dependency management implemented in CMake, primarily relying on the FetchContent module (CMake v3.11) which is capable of making the build of the subproject part of the SDK's buid (in the case of non-header-only deps). This initial work involved only the deps requried for the console samples only.

This PR allows users to self-host their dependencies (eg. using package managers) and if something is not found, they are auto-fetched by the SDK scripts.

Sample out-of-the-box experience for users not providing any deps:
```
cmake.exe -DBUILD_DOCS=OFF -DBUILD_TESTING=OFF -DOPENCL_SDK_BUILD_SAMPLES=ON -DOPENCL_SDK_BUILD_OPENGL_SAMPLES=OFF -Hc:/Users/mate/Source/Repos/OpenCL-SDK -Bc:/Users/mate/Source/Repos/OpenCL-SDK/.vscode/build -G "Ninja Multi-Config"
...
[cmake] -- cargs (https://github.com/likle/cargs) not found. To self-host, set cargs_INCLUDE_PATH and cargs_LIBRARY to point to the headers and library respectively adding '-D cargs_INCLUDE_PATH=/path/to/cargs/include/dir -D cargs_LIBRARY/path/to/cargs/libcargs' to the cmake command. (missing: cargs_INCLUDE_PATH cargs_LIBRARY) 
[cmake] -- Fetching cargs.
[cmake] -- Adding cargs subproject: C:/Users/mate/Source/Repos/OpenCL-SDK/.vscode/build/_deps/cargs-external-src
[cmake] -- TCLAP (http://tclap.sourceforge.net/) not found. To self-host, set TCLAP_INCLUDE_PATH to point to the headers adding '-DTCLAP_INCLUDE_PATH=/path/to/tclap' to the cmake command. (missing: TCLAP_INCLUDE_PATH) 
[cmake] -- Fetching TCLAP.
[cmake] -- Found TCLAP: C:/Users/mate/Source/Repos/OpenCL-SDK/.vscode/build/_deps/tclap-external-src/include  
[cmake] -- Stb (https://github.com/nothings/stb) not found. To self-host, set Stb_INCLUDE_PATH to point to the headers adding '-D Stb_INCLUDE_PATH=/path/to/stb' to the cmake command. (missing: Stb_INCLUDE_PATH) 
[cmake] -- Fetching Stb.
[...]

PR #42, opened 2022-03-01
Auto-fetch missing dependencies

MathiasMagnus:
This PR supersedes #40 as it entails both console and GUI dependency fetching. This PR has few warning fixes, ABI fixes (missing exports, includes using OpenCLSDK.lib) and shorter, cleaner history. As a consequence, they don't apply on top of each other.


bashbaug (approved):
This is great!

I didn't do a complete review, especially for the CI changes, but I did try building on both Windows and Linux and it seemed to "just work" - excellent.

There are a few warnings in the Windows build that would be nice to eliminate sometime, but they can be fixed in a separate PR.

Could/should any of the README docs be simplified as a result of these changes?  In particular, should we still recommend using vcpkg?

PR #43, opened 2022-03-09
Initial release scripts

MathiasMagnus:
This PR holds automation for source and binary releases of the SDK whenever a version tag is applied. It only creates a draft release which then has to be manually verified and published. (The UI is fairly straightforward) A draft release which runs after adding a version tag look like [this](https://github.com/StreamHPC/OpenCL-SDK/releases/tag/untagged-1c69145d6331af93ae5b).

Currently only Windows binary releases are done, because that is the platform which necessitates this the most, but later iterations can expand on that. (I would call on people actually shipping Linux releases, not just libraries to tune in on what is the optimal set of release build flags on Linux.) Submodule source files are included in the source tarballs.

_(Unfortunately I can't remove the automatic source tarball assets, those are always added by GitHub and can't be removed. I already [reached out to the community](https://github.community/t/remove-overwrite-default-source-tarball-assets/235118) if that's possible, but that is more or less polish atop the feature. It's important, we don't want people downloading the wrong stuff, but there's really no way around it, or not that I know of.)_

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@ cmake_minimum_required(VERSION 3.0)
 set(CMAKE_CXX_STANDARD 14)
 
 project(OpenCL-SDK
-  VERSION 1.0
+  VERSION 2022.03.09
   LANGUAGES
     C CXX
 )
@@ -102,3 +102,5 @@ install(
   FILES ${CMAKE_CURRENT_BINARY_DIR}/OpenCL/OpenCLConfigVersion.cmake
   DESTINATION ${config_package_location}
 )
+
+include(CPack)
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,133 @@
+on:
+  push:
+    tags:
+      - 'v*'
+
+name: Create Release
+
+jobs:
+  windows-binary:
+    runs-on: windows-2022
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ cmake_minimum_required(VERSION 3.0)
 set(CMAKE_CXX_STANDARD 14)
 
 project(OpenCL-SDK
-  VERSION 1.0
+  VERSION 2022.03.09
```
bashbaug, line 20: Presumably this needs to change when we do a release?  Does anything else need to change?

I think it would be useful to have a short document that describes the process for doing a release.

bashbaug (approved):
This is cool!

I haven't tried this but all of the steps in the script look reasonable.  I like how it creates a draft release by default so we still need a small amount of manual checking before publishing a release.

It would be helpful to have a short document that enumerates the steps needed to perform a release.  It's fine if it's brief for now; we can always refine later.

CMakeLists.txt
MathiasMagnus: I've added such a document. Please check if it's in line with what you had in mind.

bashbaug: This is perfect - thanks!

bashbaug (approved):

PR #46, opened 2022-05-12
Fix utility library install

MathiasMagnus:
Because there are no consumption tests for the utility libraries (yet), it went unnoticed that consumers of the SDK can't consume the utility libraries using the advertized name `OpenCL::Utils` and `OpenCL::UtilsCpp`, but instead the exported names were left to be the default (target name), hence the exported names turned out to be `OpenCL::OpenCLUtils` etc.

It also went unnoticed, that consumers of SDK installs were missing the config header of the utility libraries from the install tree.

This PR fixes both issues, falling under the same umbrella.


bashbaug (approved):
LGTM, no reason to wait to merge this.

PR #50, opened 2022-09-13
Fix UtilsCpp debug install

MathiasMagnus:
On Windows, Debug and Release C++ libraries aren't link compatible. If the name of a library on disk is the same for both Debug and Release builds, they can't be installed side-by-side and makes it much harder to consume them for development purposes where both versions would be needed. This PR addresses that by changing the name of the utility libraries on disk for Debug builds, so they can coexist with Release artifacts.

_(Note that CI doesn't test it yet, because the OpenCL-SDK's CI is still fairly rudimentary. This feature still requires manual testing.)_

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -83,10 +83,9 @@ jobs:
         shell: pwsh
         run: |
           if ('${{ matrix.deps }}' -eq 'vcpkg') {
-            git clone https://github.com/Microsoft/vcpkg.git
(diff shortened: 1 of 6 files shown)


bashbaug (changes requested):
Mostly LGTM, I'm able to build the utils lib with these changes and I see the "d" suffix on the debug lib.  There is one issue in the .cpp test that needs to be fixed before merging, though.

bashbaug (approved):
Latest version LGTM - thanks!

PR #51, opened 2022-09-13
Fix GLEW detection for FetchContent path

MathiasMagnus:
Because we don't thoroughly test fetching dependencies (as we state it's on a "best effort basis), it flew under the radar, that GLEW when autofetched will not fire the predicate it's supposed to.


bashbaug (approved):
Works for me, didn't manually test but CI is passing.

PR #52, opened 2022-09-13
Ability to pause sim

MathiasMagnus:
When I wanted to create screencaptures of these samples, it proved useful to start/stop them on demand. In those cases I changed the starting value of the bool controlling the animation to `false`, but that may confuse people who just launch the sample, so the default is now `true`.


bashbaug (approved):
LGTM!

PR #59, opened 2022-09-26
Release fixes

### LINKED ISSUE #57: Inconsistent version naming scheme in OpenCL-SDK-vJJJJ.MM.DD-Source.tar.gz
In the newest release, the downloadable source file OpenCL-SDK-v2022.09.23-Source.tar.gz has the directory OpenCL-SDK-2022.5.18-Source.
Not only do the dates not match up, but the version string of the directory doesn't contain a 0 in the first space, when the day or the month is less than 10.

This inconsistency makes it quite hard to automate things.

MathiasMagnus:
Fixes #57

I fixed the swallowing of leading zeroes by setting the [CMP0096](https://cmake.org/cmake/help/latest/policy/CMP0096.html) to `NEW`, I also removed manual renaming of the package files and instead give them proper names through CPack. I also added a check to make sure that the project version matches the tag name. Sample release looks like [this](https://github.com/StreamHPC/OpenCL-SDK/releases/tag/untagged-039fd188fc39159408bb) using this CD script, when the project version mismatches the tag name, the script fails like [this](https://github.com/StreamHPC/OpenCL-SDK/actions/runs/3130789378/jobs/5081426187).

Feel free to squash.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,24 +48,14 @@ jobs:
           -D OPENCL_SDK_BUILD_SAMPLES=OFF `
           -D OPENCL_ICD_LOADER_BUILD_TESTING=OFF
 
-    - name: Build
+    - name: Check version number
       shell: pwsh
       run: |
-        & cmake `
-          --build "${env:GITHUB_WORKSPACE}\build" `
-          --config Release `
-          -- `
-          /verbosity:minimal `
-          /maxCpuCount `
-          /noLogo
-
-    - name: Install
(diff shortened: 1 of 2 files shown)


bashbaug (approved):
Thanks for the quick turnaround!

It would be nice to drive the filename exclusively from the git tag to eliminate the possibility of a mismatch but at least this will prevent releases with the mismatch.

I think we'll want another release ASAP, in which case we'll need to bump the project version again, but we can do this via a separate PR.

PR #65, opened 2023-02-15
Add C++ variant for binaries sample

Dastin-Wilski-Mobica:
* file utils were expanded to support reading and writing binaries

### DIFF
--- a/lib/src/Utils/File.cpp
+++ b/lib/src/Utils/File.cpp
@@ -1,0 +1,119 @@
+// OpenCL SDK includes
+#include <CL/Utils/File.hpp>
+
+// STL includes
+#include <fstream>
+#include <iterator>
+#include <algorithm>
+#include <iostream>
+
+std::string cl::util::read_text_file(const char* const filename,
+                                     cl_int* const error)
+{
+    std::ifstream in(filename);
+    if (in.good())
+    {
+        try
+        {
+            std::string red((std::istreambuf_iterator<char>(in)),
+                            std::istreambuf_iterator<char>());
+
+            *error = CL_SUCCESS;
+            return red;
+
+        } catch (std::bad_alloc& ex)
+        {
+            (void)ex;
+            detail::errHandler(CL_OUT_OF_RESOURCES, error, "Bad allocation!");
+            return std::string();
+        }
+    }
+    else
+    {
+        detail::errHandler(CL_INVALID_VALUE, error, "No file!");
+        return std::string();
+    }
+}
+
+std::vector<unsigned char>
+cl::util::read_binary_file(const char* const filename, cl_int* const error)
+{
+    std::ifstream in(filename, std::ios::binary);
+    if (in.good())
+    {
+        try
+        {
+            std::vector<unsigned char> buffer(
+                std::istreambuf_iterator<char>(in), {});
+            return buffer;
+        } catch (std::bad_alloc& ex)
+        {
+            (void)ex;
+            detail::errHandler(CL_OUT_OF_RESOURCES, error, "Bad allocation!");
+            return std::vector<unsigned char>();
+        }
+    }
+    else
+    {
+        *error = CL_INVALID_VALUE;
+        return std::vector<unsigned char>();
+    }
+}
+
+
+cl::Program::Binaries
+cl::util::read_binary_files(const std::vector<cl::Device>& devices,
+                            const char* const program_file_name,
+                            cl_int* const error)
+{
+    cl::Program::Binaries binaries(0);
+    size_t num_devices = devices.size();
+
+    for (const auto& device : devices)
+    {
+        string device_name = device.getInfo<CL_DEVICE_NAME>();
+        string binary_name =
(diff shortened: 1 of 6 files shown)


samples/core/binaries/main.cpp
```diff
@@ -0,0 +1,217 @@
...
+ */
+
+// OpenCL SDK includes
+#include <CL/Utils/Utils.hpp>
+#include <CL/SDK/Context.hpp>
+#include <CL/SDK/Options.hpp>
+#include <CL/SDK/CLI.hpp>
+
+// STL includes
+#include <iostream>
```
bashbaug, line 24: The CI builds are failing on Linux due to a missing include:

```suggestion
#include <climits>
#include <iostream>
```

bashbaug (changes requested):

lib/src/Utils/File.cpp
```diff
@@ -0,0 +1,119 @@
...
+    {
+        *error = CL_INVALID_VALUE;
+        return std::vector<unsigned char>();
+    }
+}
+
+
+cl::Program::Binaries
+cl::util::read_binary_files(const std::vector<cl::Device>& devices,
+                            const char* const program_file_name,
```
MathiasMagnus, line 66: If program_file_name is only a base file name, perhaps we should indicate that in the variable name. IntelliSense will confuse people with this variable name _and_ no Doxygen summary popping up.

lib/src/Utils/File.cpp
```diff
@@ -0,0 +1,119 @@
...
+    for (const auto& device : devices)
+    {
+        string device_name = device.getInfo<CL_DEVICE_NAME>();
+        string binary_name =
+            string(program_file_name) + "-" + device_name + ".bin";
+
+
+        binaries.push_back(
+            cl::util::read_binary_file(binary_name.c_str(), error));
+        if (*error == CL_INVALID_VALUE) return binaries;
```
MathiasMagnus, line 81: I think `errHandler` is missing here too.

lib/src/Utils/File.cpp
```diff
@@ -0,0 +1,119 @@
...
+        } catch (std::bad_alloc& ex)
+        {
+            (void)ex;
+            detail::errHandler(CL_OUT_OF_RESOURCES, error, "Bad allocation!");
+            return std::vector<unsigned char>();
+        }
+    }
+    else
+    {
+        *error = CL_INVALID_VALUE;
```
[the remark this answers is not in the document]
MathiasMagnus: Also, the error is never checked for being a `nullptr`. Check before assigning.

samples/core/binaries/main.cpp
```diff
@@ -0,0 +1,218 @@
...
+                      << "Selected device: "
+                      << devices.at(0).getInfo<CL_DEVICE_NAME>() << "\n"
+                      << std::endl;
+        }
+
+        /// Try to read binary
+        cl::Program::Binaries binaries =
+            cl::util::read_binary_files(devices, "Collatz", &error);
+
+        if (error != CL_SUCCESS)
```
MathiasMagnus, line 96: All samples are built with exception handling enabled. If error handling is implemented using `errHandler`, then in the presence of exceptions, trying to read the binaries which aren't on disk yet result in an exception. As such, compiling the binaries should be in a `catch(cl::util::Error&)` clause, so if any of the files did not exist on disk, then we compile from scratch and save.

MathiasMagnus: As a sidenote, the utility library could expose a kernel caching mechanism of some sort. Either using the current working directory, or some other more sophisticated storage mechanism. Other libraries roll their own kernel cache (clFFT for instance), but this is a recurring issue and could be captured by a dedicated utility. This is nothing to act upon now, but we should capture in an issue.

MathiasMagnus (changes requested):
Nice addition. Some changes are requested to align better with existing code and behavior.

samples/core/binaries/main.cpp
Dastin-Wilski-Mobica: Thanks for review!

I've tried to address your comments in the following commit.
I hope its better now.

MathiasMagnus (changes requested):
Good changes. Only one error-path seems to omit exception behavior.

lib/src/Utils/File.cpp
```diff
@@ -45,42 +44,45 @@ cl::util::read_binary_file(const char* const filename, cl_int* const error)
...
 
         binaries.push_back(
             cl::util::read_binary_file(binary_name.c_str(), error));
-        if (*error == CL_INVALID_VALUE) return binaries;
+        if (error != nullptr)
+        {
+            if (*error != CL_SUCCESS)
+            {
+                return cl::Program::Binaries();
+            }
```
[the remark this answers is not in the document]
Dastin-Wilski-Mobica: done

MathiasMagnus (approved):
👍🏻

bashbaug (approved):
LGTM and the Linux builds are fixed - thanks!

PR #66, opened 2023-02-19
add the OpenCL Extension Loader to the OpenCL SDK

bashbaug:
The OpenCL Extension Loader provides definitions for OpenCL extension APIs which allows  them to be called like any other OpenCL API functions.  The OpenCL extension loader handles querying OpenCL API functions and managing different sets of OpenCL extension functions for different OpenCL platforms.

Please note that the outbound license for the OpenCL Extension Loader is dual-license Apache 2.0 (consistent with this repo) and MIT.

The files in this PR are taken unchanged from my [opencl-extension-loader repo](https://github.com/bashbaug/opencl-extension-loader) to allow for easy updates and sync'ing in the future.


bashbaug:
I'd prefer not to reformat this code because it is automatically generated.  Let me know how you'd like to handle this.  One easy option would be to add `/* clang-format off */` to the generated files.

MathiasMagnus:
This looks like an awesome addition. My first question would be: would consuming the source code in a snapshot be the best way of the SDK consuming the extension loader? Almost every other component is a Git submodule. I'm uncertain if taking semi-frequent snapshots of the upstream repo is the easiest to maintain.

I don't expect a lot of friction from the code quality perspective. It look top notch, the install definition looks almost as if it always targeted to be consumed by the SDK.

bashbaug:
> would consuming the source code in a snapshot be the best way of the SDK consuming the extension loader?

Good question.  Couple of options:

- Once the extension loader is in the OpenCL SDK there's a very real chance I can sunset my repo and then the OpenCL SDK will become the single "source of truth".  This seems like the best option.
- We could define a new KhronosGroup repo for it and then pull it in via a submodule, but I don't think that buys us much unless we have plans for other loader-like libraries such as a "[runtime loader](https://github.com/bashbaug/opencl-runtime-loader)".
- We put it into an existing repo and pull it in via a submodule.  The OpenCL ICD loader is probably the best fit here, but it's not great, so I'd prefer not to do this.

What do you think?

MathiasMagnus (approved):
Few minor bulletproofing is needed, but nothing major enough to hold back the feature.

PR #67, opened 2023-02-25
add pure C and C++ versions of the enumopencl sample

bashbaug:
This change switches the existing enumopencl sample to use pure C, compilable with a C compiler (no C++).

It also adds an equivalent C++ sample that makes extensive use of the OpenCL C++ bindings.

The output from both samples is identical.

### DIFF
--- a/samples/core/enumopencl/CMakeLists.txt
+++ b/samples/core/enumopencl/CMakeLists.txt
@@ -16,4 +16,4 @@ add_sample(
     TEST
     TARGET enumopencl
     VERSION 120
-    SOURCES main.cpp)
+    SOURCES main.c)
--- a/samples/core/enumopencl/CMakeLists.txt
+++ b/samples/core/enumopencl/CMakeLists.txt
@@ -17,3 +17,10 @@ add_sample(
     TARGET enumopencl
     VERSION 120
     SOURCES main.c)
+
+add_sample(
+    TEST
+    TARGET enumopenclcpp
+    VERSION 120
+    SOURCES main.cpp)
+    
--- a/samples/core/enumopencl/CMakeLists.txt
+++ b/samples/core/enumopencl/CMakeLists.txt
@@ -23,4 +23,3 @@ add_sample(
     TARGET enumopenclcpp
     VERSION 120
     SOURCES main.cpp)
-    
--- a/samples/core/enumopencl/main.c
+++ b/samples/core/enumopencl/main.c
@@ -85,10 +85,6 @@ static cl_int PrintPlatformInfoSummary(cl_platform_id platformId)
     free(platformVendor);
     free(platformVersion);
 
-    platformName = NULL;
-    platformVendor = NULL;
-    platformVersion = NULL;
-
     return errorCode;
 }
 
@@ -194,12 +190,6 @@ static cl_int PrintDeviceInfoSummary(cl_device_id* devices, cl_uint numDevices)
         free(deviceVersion);
         free(deviceProfile);
         free(driverVersion);
-
-        deviceName = NULL;
-        deviceVendor = NULL;
-        deviceVersion = NULL;
-        deviceProfile = NULL;
-        driverVersion = NULL;
     }
 
     return errorCode;
--- a/samples/core/enumopencl/main.c
+++ b/samples/core/enumopencl/main.c
@@ -15,25 +15,18 @@
  */
 
 #include <stdio.h>
-#include <vector>
+#include <stdlib.h>
 
 #include <CL/cl.h>
 
 static cl_int AllocateAndGetPlatformInfoString(cl_platform_id platformId,
                                                cl_platform_info param_name,
-                                               char*& param_value)
+                                               char** pStr)
 {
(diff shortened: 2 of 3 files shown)


samples/core/enumopencl/CMakeLists.txt
```diff
@@ -16,4 +16,11 @@ add_sample(
...
     TARGET enumopencl
     VERSION 120
+    SOURCES main.c)
+
+add_sample(
+    TEST
+    TARGET enumopenclcpp
+    VERSION 120
     SOURCES main.cpp)
+
```
keryell, line 26: EOL

samples/core/enumopencl/main.cpp
```diff
@@ -17,132 +17,18 @@
 #include <stdio.h>
```
keryell, line 17: iostream in C++?

samples/core/enumopencl/main.cpp
```diff
@@ -17,132 +17,18 @@
...
-
-    if (errorCode != CL_SUCCESS)
-    {
-        delete[] param_value;
-        param_value = NULL;
-    }
-
-    return errorCode;
+    printf("\tName:           %s\n",
+           platform.getInfo<CL_PLATFORM_NAME>().c_str());
```
keryell, line 25:
```suggestion
    std::cout << "\tName:           " << platform.getInfo<CL_PLATFORM_NAME>();
```
and put the `\n` before the next `\t`.

samples/core/enumopencl/main.cpp
```diff
@@ -155,117 +41,44 @@ static void PrintDeviceType(const char* label, cl_device_type type)
            (type & CL_DEVICE_TYPE_CUSTOM) ? "CUSTOM " : "");
 }
 
-static cl_int PrintDeviceInfoSummary(cl_device_id* devices, size_t numDevices)
+static cl_int PrintDeviceInfoSummary(const std::vector<cl::Device> devices)
```
keryell, line 44:
```suggestion
static cl_int PrintDeviceInfoSummary(const std::vector<cl::Device>& devices)
```

bashbaug:
Thanks @keryell , I think I fixed all of these, except I didn't combine the previous `\n` with the current `\t` just because it seemed a little too clever for this simple sample.

keryell (approved):
Thanks!

samples/core/enumopencl/main.c
```diff
@@ -0,0 +1,243 @@
...
+        free(deviceVendor);
+        free(deviceVersion);
+        free(deviceProfile);
+        free(driverVersion);
+
+        deviceName = NULL;
+        deviceVendor = NULL;
+        deviceVersion = NULL;
+        deviceProfile = NULL;
+        driverVersion = NULL;
```
MathiasMagnus, line 202: Is this idiomatic C to set function local variable to `NULL` after having called `free()` on them?

samples/core/enumopencl/main.c
```diff
@@ -0,0 +1,243 @@
...
+
+        cl_device_id* devices =
+            (cl_device_id*)malloc(numDevices * sizeof(cl_device_id));
+        clGetDeviceIDs(platforms[i], CL_DEVICE_TYPE_ALL, numDevices, devices,
+                       NULL);
+
+        PrintDeviceInfoSummary(devices, numDevices);
+        printf("\n");
+
+        free(devices);
```
MathiasMagnus, line 235: Even though devices obtained by `clGetDeviceIDs` will return "root-level" devices which never get freed, I would highly suggest promoting "brainlessly" calling `clReleaseDevice()` on anything that was created at scope-level and then the scope closes without transferring ownership of those devices.

It's just a hygenic thing, much like freeing memory right before main exits. It's not strictly required, the OS will clean up anyway, but the reflex should develop in people to control the lifetime of _all_ OpenCL objects that have retain/release semantics.

MathiasMagnus (changes requested):
Nice addition. Minor change request.

samples/core/enumopencl/main.c
bashbaug: This was a habit I got into to try to avoid possible double-frees, but it's unnecessary for this sample, and I wasn't doing it consistently anyhow.  Removed.

samples/core/enumopencl/main.c
bashbaug: I think I'd prefer not to do this.  Here's why:

- Most uses of device handles will not create sub-devices, so most uses will not need to release device handles.  I'd prefer not to put extranenous code in such a simple sample.
- clReleaseDevice is an OpenCL 1.2 function, so it may behave unpredictably for an OpenCL 1.0 or 1.1 implementation.

Are you OK leaving this as-is?

MathiasMagnus: Okay, this isn't the hill I'd want to die on. I'd air on the side of building fail-safe muscle reflexes instead of inviting discussion on why omitting this is valid in this particular case.

MathiasMagnus (approved):

PR #71, opened 2023-03-21
update project version and bump submodules

Mikolaj-Filar-Mobica:
I'm not sure if I'm allowed to bump submodules, but it would be useful for the _blurcpp_ example I want to commit.
It updates:

1. OpenCL-CLHPP
2. OpenCL-Headers
3. OpenCL-ICD-Loader

to the latest main.


bashbaug:
Discussed on the March 28th teleconference.  Will merge as soon as CI passes.

PR #74, opened 2023-04-11
Reduce warning in OpenCL SDK

pwisniewskimobica:
I reduced the warnings by doing small changes in File.c
Mostly it was changing `fopen` occurances to `fopen_s`

### DIFF
--- a/lib/src/Utils/File.c
+++ b/lib/src/Utils/File.c
@@ -34,11 +34,19 @@ char *cl_util_read_text_file(const char *const filename, size_t *const length,
         }                                                                      \
     } while (0)
 
+#ifndef _WIN32
+#define fopen_s(fp, fmt, mode)                                                 \
+    ({                                                                         \
+        *(fp) = fopen((fmt), (mode));                                          \
+        (*(fp)) ? 0 : -1;                                                      \
+    })
+#endif
+
     /* File name can not be NULL. */
     IF_ERR(!filename, CL_INVALID_ARG_VALUE, end);
 
     /* Open file. */
-    IF_ERR(!(in = fopen(filename, "r")), CL_INVALID_VALUE, end);
+    IF_ERR((fopen_s(&in, filename, "r") != 0), CL_INVALID_VALUE, end);
 
     /* A read error already occurred? */
     IF_ERR(ferror(in), CL_UTIL_FILE_OPERATION_ERROR, fl);
@@ -116,7 +124,7 @@ unsigned char *cl_util_read_binary_file(const char *const filename,
     IF_ERR(!filename, CL_INVALID_ARG_VALUE, end);
 
     /* Open file. */
-    IF_ERR(!(in = fopen(filename, "rb")), CL_INVALID_VALUE, end);
+    IF_ERR((fopen_s(&in, filename, "rb") != 0), CL_INVALID_VALUE, end);
 
     /* A read error already occurred? */
     IF_ERR(ferror(in), CL_UTIL_FILE_OPERATION_ERROR, fl);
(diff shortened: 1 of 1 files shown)


bashbaug (changes requested):
This doesn't work on Linux, unfortunately:

```sh
/home/bashbaug/git/opencl-sdk/lib/src/Utils/File.c:41:13: warning: implicit declaration of function ‘fopen_s’; did you mean ‘fopen’?
```

A typical way to fix warnings like these is to just define `_CRT_SECURE_NO_WARNINGS` for Windows builds.  See, for example:

https://github.com/KhronosGroup/OpenCL-CTS/blob/20afedbd4a656a36132bfc42bd7519c426a1fa49/CMakeLists.txt#L153

bashbaug:
Alternatively, if we really want to keep the "secure" versions, we could either:

1. Add ifdefs and use `fopen_s` on Windows and `fopen` otherwise.
2. Add wrapper functions (maybe even a macro?) that calls `fopen_s` on Windows and `fopen` otherwise.

MathiasMagnus:
I would be highly against using `_CRT_SECURE_NO_WARNINGS`. It may pass in the CTS (even there it's not a good way), but the utility libraries are meant to be just as infrastructure as the rest of the ecosystem, encouraging downstreams to consume it. I wouldn't want ultrasound machines or whatever to ship code with security warnings disabled. That is not how we should roll.

_(And yes, secure versions of CRT functions is a mess in ISO C. Microsoft implements them, but fail to define the macro that would allow detecting their presence. GCC has POSIX or vendor-specific solution pre-dating secure CRT and have no intention implementing them. Hooray!)_

bashbaug:
FWIW, `fopen_s` is [supposed to sometimes be in C11](https://en.cppreference.com/w/c/io/fopen), but I tried enabling C11 compilation for the OpenCL SDK...

```
set(CMAKE_C_STANDARD 11)
```

... and I still ended up with linker errors on Linux.

I think a wrapper function/macro for non-Windows compiles is going to be our best bet.  We could possibly refine this further based on things like `__STDC_LIB_EXT1__`, but I'm not sure if it's worth the work.

pwisniewskimobica:
@bashbaug I just recommit this.
With creating additional function like fopen_no_warnings() independable from platform there will be more changes in the code- thats why I propose this solution with #ifdef's
What you think?

bashbaug (changes requested):
This is getting better but there is a new issue now.

We could also look at defining a wrapper or macro once, at the top of this file, vs. introducing multiple #ifdefs.

https://stackoverflow.com/questions/1513209/is-there-a-way-to-use-fopen-s-with-gcc-or-at-least-create-a-define-about-it

pwisniewskimobica:
@bashbaug Thanks for comments
I applied one of stack overflow solututions.
-->done

MathiasMagnus:
👍🏻 If it resembles the original enough, please attribute according to CC license. (refer to the reading a file implementation attributed in a CC-compliant fashion.)

pwisniewskimobica:
@MathiasMagnus 
CC license comment introduced
Thanks

pwisniewskimobica:
@MathiasMagnus @bashbaug 
A bit different define introduced.
Please take a look

MathiasMagnus (approved):
LGTM

MathiasMagnus:
MacOS is failing. `__unix` only is not a good descriminator.

pwisniewskimobica:
@MathiasMagnus in newest version is "#ifndef _WIN32"
CI passing for all configurations .
Regards

PR #77, opened 2023-04-24
Executable folder utilities

MathiasMagnus:
## Motivation

One of the annoying aspects of OpenCL/GL, Vulkan, etc. is the fact that locating assets in general on disk is a pain. Of course large projects solve this using some utility, but toy projects that are still enjoying their dependency-free world are stuck with what the CRT/STL provides if they wish to be portable (a bit like SDK samples).

`fopen` and `ifstream` both interpret the paths relative to the current working directory (CWD), which is unfortunate, because execution success now depends on launching our code from a specific location our code assumes. This assumption definitely blows up in unsuspecting users' faces. C++ luckily has a passable solution: `std::filesystem::canonical(argv[0])` which by some magical alignment [happens to works on all compilers](https://godbolt.org/z/xPMceanbW) (CE recently broke execution for MSVC compilers, so those without a working Win env just have to trust me on that it works). Note that even this isn't ideal, because one has to relay the `argv` symbol all the way to where this translation occurs. In sample code, that's okay, but more complex programs want `argv` out of thin air. ([D2567](https://isocpp.org/files/papers/D2567R0.pdf) targets just this, so maybe in C26 🤞🏻) A mildly elegant solution would be [#embed in C23](https://thephd.dev/finally-embed-in-c23), but mandating a C23 compiler in 2023 is reckless at best. Unfortunately C18 and under doesn't have a similar feature in the CRT, therefore our C samples are subject to CWD relativeness. Seeing this blow up in my students' (and co-lecturers') faces makes me sad. 😞

## Proposed solution

This problem needs a solution. I cooked up a utility for `OpenCLUtils.lib`. It follows the idiom of `clGetXyzInfo()` functions, first allocate, then return value. It's not dependent on `argv` so one may invoke it in the depths of any library init. It wraps [whereami](https://github.com/gpakosz/whereami), a project with the most permissive dual-license to date: [MIT](https://github.com/gpakosz/whereami/blob/master/LICENSE.MIT)/[WTFPLv2](https://github.com/gpakosz/whereami/blob/master/LICENSE.WTFPLv2). The use of this lib is an implementation detail, it does not leak to consumers of the utility library.

The samples in my opinion became much nicer and now they don't fail, regardless of where they're launched from.

## Changes introduced

- Added the previously mentioned utility libraries
- Updated all samples using kernels to use them
- Add an overdue option `OPENCL_SDK_BUILD_UTILITY_LIBRARIES` CMake option
  - As courtesy to those not needing them, especially how the non-SDK utility picked up it's 1st dependency
- **Breaking change:** added default arguments to all C++ file utility functions. I believe it was a bug that these default args were absent. Intention was to follow the behavior of `opencl.hpp` in all regards and make a utility library that is a natural extension to that.


lib/Utils.md
```diff
@@ -119,3 +120,97 @@ public:
...
+    cl_int* const error);
+```
+
+```c++
+std::string read_exe_relative_text_file(
+    const char* const filename,
+    cl_int* const error = nullptr);
+```
+
+These functions read a text file into memory, where `filename` is evaluated relative to the executable currently running. The C-version contains a terminating null and takes an optional pointer to `length` by which the length my be returned, potentially saving a subsequent call to `strlen`. The function hands ownership of the allocated storage to the caller.
```
keryell, line 216:
```suggestion
These functions read a text file into memory, where `filename` is evaluated relative to the executable currently running. The C-version contains a terminating null and takes an optional pointer to `length` by which the length will be returned, potentially saving a subsequent call to `strlen`. The function hands ownership of the allocated storage to the caller.
```
It looks like a typo which has been copy-pasted everywhere.

lib/src/Utils/File.c
```diff
@@ -329,3 +332,76 @@ cl_program cl_util_read_binaries(const cl_context context,
...
+                                          size_t *const length,
+                                          cl_int *const error)
+{
+    char *result = NULL;
+    size_t result_size = 0;
+    char *abs_path = NULL;
+    cl_int err = CL_SUCCESS;
+    size_t exe_folder_length;
+    OCLERROR_RET(cl_util_executable_folder(NULL, &exe_folder_length), err, end);
+    MEM_CHECK(abs_path = (char *)malloc(sizeof(char) * exe_folder_length
```
keryell, line 392:
```suggestion
    MEM_CHECK(abs_path = (char *)malloc(exe_folder_length
```
This is 1 by definition of a `char`.

samples/extensions/khr/nbody/main.cpp
```diff
@@ -155,10 +155,8 @@ void NBody::initializeGL()
     if (glewInit() != GLEW_OK) std::exit(EXIT_FAILURE);
 
     auto create_shader = [](std::string file_path, cl_GLenum shader_stage) {
-        std::ifstream shader_stream(file_path);
-        std::string shader_string{ std::istreambuf_iterator<GLchar>{
-                                       shader_stream },
-                                   std::istreambuf_iterator<GLchar>{} };
+        std::string shader_string =
+            cl::util::read_exe_relative_text_file(file_path.c_str());
```
keryell, line 159: Add an overload taking a `std::string`. Actually a `string_view` would be even better when it is available.

keryell (changes requested):
That looks like a good improvement.

cmake/Dependencies/whereami/whereami.cmake
```diff
@@ -0,0 +1,15 @@
...
+  GIT_TAG        ba364cd54fd431c76c045393b6522b4bff547f50 # master @ 2023.04.20.
+)
+FetchContent_MakeAvailable(whereami-external)
+add_library(whereami IMPORTED INTERFACE)
+target_include_directories(whereami
+  INTERFACE "${CMAKE_CURRENT_BINARY_DIR}/_deps/whereami-external-src/src"
+)
+target_sources(whereami
+  INTERFACE "${CMAKE_CURRENT_BINARY_DIR}/_deps/whereami-external-src/src/whereami.c"
+)
```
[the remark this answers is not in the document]
MathiasMagnus: Done.

samples/extensions/khr/nbody/main.cpp
MathiasMagnus: I would like to make that a separate discussion, because some of the overloads would need to be version guarded.

lib/Utils.md
```diff
@@ -119,3 +120,97 @@ public:
...
+    cl_int* const error);
+```
+
+```c++
+std::vector<unsigned char> read_binary_file(
+    const char* const filename,
+    cl_int* const error = nullptr);
+```
+
+These functions read a binary file into memory, where `filename` is evaluated relative to the current working directory. The C-version takes an optional pointer to `length` by which the length my be returned. Beause it's binary data, it is _not_ null-terminated, therefore highly recommended to take it's size. The returned types align with OpenCL APIs taking binaries as input. The function hands ownership of the allocated storage to the caller.
```
bashbaug, line 154:
```suggestion
These functions read a binary file into memory, where `filename` is evaluated relative to the current working directory. The C-version takes an optional pointer to `length` by which the length my be returned. Because it's binary data, it is _not_ null-terminated, therefore highly recommended to take it's size. The returned types align with OpenCL APIs taking binaries as input. The function hands ownership of the allocated storage to the caller.
```

samples/extensions/khr/nbody/main.cpp
bashbaug: I agree it would be cleaner if the C++ version had a `const std::string&` overload, which shouldn't need any version guards.  This could be done in a subsequent PR, though.

samples/extensions/khr/nbody/main.cpp
```diff
@@ -216,8 +214,8 @@ void NBody::initializeGL()
         return program;
     };
 
-    vertex_shader = create_shader("./nbody.vert.glsl", GL_VERTEX_SHADER);
-    fragment_shader = create_shader("./nbody.frag.glsl", GL_FRAGMENT_SHADER);
+    vertex_shader = create_shader("nbody.vert.glsl", GL_VERTEX_SHADER);
+    fragment_shader = create_shader("nbody.frag.glsl", GL_FRAGMENT_SHADER);
```
bashbaug, line 218: Minor, will this `./filename.cl` syntax still work?  Or is it just unnecessary now, so no reason to include it?

bashbaug (approved):
Found one typo, otherwise LGTM.

bashbaug:
@keryell can you please check if your comments have been addressed?  We'd like to merge this shortly - thanks!

keryell (approved):
Thanks.

samples/extensions/khr/nbody/main.cpp
MathiasMagnus: It should. The implementation gets the absolute path to the executable's folder and appends a `/` unconditionally before concatenating it with the user-provided string. `"/home" + "/" + "./kernel.cl"` should work. It's not a normalized path, but it evaluates to the same thing without the `./`

bashbaug:
Not sure what's caused the issue with the Windows builds but they do appear to be "real":

> D:\a\OpenCL-SDK\OpenCL-SDK\samples\extensions\khr\conway\main.cpp(51,26): error C2397: conversion from 'sf::Style::<unnamed-enum-None>' to 'sf::Uint32' requires a narrowing conversion [D:\a\OpenCL-SDK\OpenCL-SDK\build\samples\extensions\khr\conway\conwaycpp.vcxproj]

MathiasMagnus:
@bashbaug @keryell I added a few fixes to the tip of the branch, partly to fix what Ben pointed out and addressed some warnings.

The `sf:::Uint32` conversion is an SFML API facepalm. `sf::Window` [takes](https://github.com/SFML/SFML/blob/0da4382d6c85409558f712f05c252a3ac275f976/include/SFML/Window/Window.hpp#L86) its style enum flag as an `sf::Uint32`, however the enum which supplies the valid values is an [_anonymous enum_](https://github.com/SFML/SFML/blob/master/include/SFML/Window/WindowStyle.hpp). Our `cl::sdk::InteropWindow` followed `sf::Window`'s suit and took the style param as an unsigned integer, but the default [underlying type](https://en.cppreference.com/w/cpp/types/underlying_type) of the anon enum is `int`, hence the narrowing conversion.

I changed the API of `cl::sdk::InteropWindow` to take the underlying type of the enum in its CTOR and converts it internally before handing it over to `sf::Window`. This API change does not leak to consumers, because the SDK lib isn't exported, there we're free to change the API.

GCC also argued about unhandled switch cases, and adding a `default` there is safe as we don't intend to handle all cases anyway, just the ones relevant to us.

bashbaug (approved):
LGTM, and all CI is passing now :+1: 

I'll just merge this.  Thanks!

PR #78, opened 2023-05-02
Add submodules flags to README.md

piotr-wozniak-mobica:


bashbaug:
I don't mind adding this but I am a little worried about complicating build steps unnnecessarily and that this documentation could easily go out-of-date.  Do we really think many OpenCL SDK builders will set any of these options to non-default values?  Should we link to the different submodule documentation instead?

piotr-wozniak-mobica:
Agreed with Ben and Closing PR.

PR #79, opened 2023-06-22
Fix warning unused parameter samples

Bogumil-Sapinski-Mobica:


samples/core/enumopencl/main.cpp
```diff
@@ -67,6 +67,9 @@ static cl_int PrintDeviceInfoSummary(const std::vector<cl::Device> devices)
 
 int main(int argc, char** argv)
 {
+    (void)argc;
+    (void**)argv;
```
alycm, line 71:
```suggestion
    (void)argv;
```

While your suggested change is not wrong just alway cast to `(void)` is the idiomatic way to make this as unused.

samples/core/enumopencl/main.cpp
```diff
@@ -67,6 +67,9 @@ static cl_int PrintDeviceInfoSummary(const std::vector<cl::Device> devices)
 
 int main(int argc, char** argv)
```
alycm, line 68:
```suggestion
int main(int, char**)
```

An alternative way to eliminate the unused warning, drop the unused names.

Do not take both of my suggestions, just one, either is okay with me.

Bogumil-Sapinski-Mobica: Done as suggested.

Bogumil-Sapinski-Mobica:
CI fails on SFML which was not changed in this PR. I kindly ask you to view and solve this issue as I do not have an access to CI process internals in order to fix it.

bashbaug (approved):
LGTM!

bashbaug:
Merging as discussed in the September 12th teleconference.  The CI failures are unrelated to these changes.

PR #83, opened 2023-09-12
Fix Vcpkg build failures in CI

MathiasMagnus:
Add installing of libxcursor-dev which newer SFML versions require from the system.


bashbaug (approved):
LGTM!

PR #87, opened 2023-10-10
Implemented Callback sample

mfep:
- Implemented sample showcasing event-based synchronization of command queues and event callbacks,
- for details, please visit the newly added README file.


samples/core/callback/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+int main(void) { thrd_sleep(&(struct timespec){.tv_nsec=1}, NULL); }
+")
+
+# Signature can be modernized at CMake version 3.25
+try_compile(
+    HAS_C11_THREADS
+    "${CMAKE_CURRENT_BINARY_DIR}"
+    SOURCES "${CMAKE_CURRENT_BINARY_DIR}/threads.c"
+    C_STANDARD 11
+    C_STANDARD_REQUIRED ON
```
bashbaug, line 29: Can we simplify this logic if we require C11?  We're already requiring C++14, so perhaps this won't be controversial.

samples/core/callback/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+endif()
+
+add_sample(
+    TEST
+    TARGET callback_cpp
+    VERSION 300
+    SOURCES main.cpp
+    KERNELS reaction_diffusion.cl)
+target_link_libraries(callback_cpp PRIVATE
+    $<TARGET_NAME_IF_EXISTS:Threads::Threads>)
```
bashbaug, line 53: Do we need this?  We already check for and require C++14.

```cmake
set(CMAKE_CXX_STANDARD 14)
```

samples/core/callback/main.c
```diff
@@ -0,0 +1,517 @@
...
+                                    cb_opts.side, iter / cb_opts.write_iter,
+                                    &state.copy_event, &state.read_event),
+                error, copy_ev);
+        }
+
+        cl_mem tmp = state.img_a;
+        state.img_a = state.img_b;
+        state.img_b = tmp;
+    }
+
```
bashbaug, line 450: I'm not exactly sure that it's at this point in the program, but the C++ version has a call to `clFinish` around here.  Do we need this in the C version also?

```c++
        // Wait for every read finished callback to fire.
        read_queue.finish();
```

samples/core/callback/main.cpp
```diff
@@ -0,0 +1,341 @@
...
+        for (std::size_t iter = 0; iter < iterations; ++iter)
+        {
+            // Enqueue the next step of the iteration and swap the source and
+            // destination images. This synchronizes with the previous copy of
+            // the input image, ensuring that the copy is finished before the
+            // current kernel overwrites the data.
+            auto compute_event = reaction_diffusion_step(
+                cl::EnqueueArgs(compute_queue, copy_event,
+                                cl::NDRange(side, side)),
+                images.read, images.write);
```
bashbaug, line 255: This isn't something that needs to be changed in this sample, but noting for completeness: this C++ sample is making extra calls that aren't present in the C sample, mostly due to how this kernel is being enqueued:

1. Building the `EnqueueArgs` makes a copy of the queue and event, which requires retaining the queue and the passed-in event, then releasing them when the `EnqueueArgs` goes out of scope.
2. The variadic template arguments for the read and write images are passed-by-value and not passed-by-reference, so there is a retain and release for them also.

I think the fix for these would go in the C++ bindings and not in the sample.

If the C++ version of the sample is slower than the C version this is probably why.

samples/core/callback/main.cpp
```diff
@@ -0,0 +1,341 @@
...
+// TCLAP includes
+#include <tclap/CmdLine.h>
+
+namespace {
+
+struct ReadJob
+{
+    // Although the buffer is not used in the callback, we keep a reference to
+    // the OpenCL object to keep it alive while the read is performed.
+    cl::Buffer buffer;
```
bashbaug, line 43: I'm not quite following why an additional reference needs to be held here.  Shouldn't it be enough that the buffer is used by the call to `clEnqueueReadBuffer`?

samples/core/callback/main.cpp
```diff
@@ -0,0 +1,341 @@
...
+        std::async([pixels = std::move(job.data), side = job.side, job_id] {
+            const cl::sdk::Image image{
+                static_cast<int>(side),
+                static_cast<int>(side),
+                static_cast<int>(sizeof(cl_uchar4)),
+                std::move(pixels),
+            };
+            const std::string filename =
+                "callbackcpp_out" + std::to_string(job_id) + ".png";
+            cl::sdk::write_image(filename.c_str(), image);
```
bashbaug, line 90: I quite like how the C sample prints the images it wrote.  This is a good way to show progress and it hints to the user to look for images that were written.  Could a similar output be added to the C++ version also?

Note: it'd be even better to print the file names for the images that are written, for both the C and C++ versions!

samples/core/callback/main.cpp
```diff
@@ -0,0 +1,341 @@
...
+            device.getInfo<CL_DEVICE_PLATFORM>()
+        }; // https://github.com/KhronosGroup/OpenCL-CLHPP/issues/150
+
+        if (!diag_opts.quiet)
+        {
+            std::cout << "Selected platform: "
+                      << platform.getInfo<CL_PLATFORM_VENDOR>() << "\n"
+                      << "Selected device: " << device.getInfo<CL_DEVICE_NAME>()
+                      << "\n"
+                      << std::endl;
```
bashbaug, line 165: I don't care too strongly about this, but if we want the C++ version to output the same information as the C version we should also print the OpenCL C version here.  See: `cl_util_print_device_info`

We could also just file an issue for this and (potentially) fix it later.

samples/core/callback/CMakeLists.txt
mfep: Unfortunately no. This very convoluted logic to determine whether the compiler supports C11 threads was required, due to the varying compiler support across the platforms we compile to.

Most notably MSVC doesn't have it, and also it cannot report C standard conformance to CMake properly (see [C_STANDARD_REQUIRED](https://cmake.org/cmake/help/latest/prop_tgt/C_STANDARD_REQUIRED.html)). This is the reason why not having C11 threads only emits a warning, and is not an error.

A proper C11 threads implementation [is coming to MSVC](https://devblogs.microsoft.com/cppblog/c11-threads-in-visual-studio-2022-version-17-8-preview-2/), but at the time of writing this code, it was not available in the CI runners, nor we can assume that everyone has that version of Visual Studio.

samples/core/callback/CMakeLists.txt
mfep: Makes sense, however some entry/entries in the supported platform/compiler matrix required an explicit linkage against pthreads. I don't remember which one was it anymore, but I can dig it up, if you wish.

samples/core/callback/main.c
mfep: Very observant finding, thank you!

Shortly: no, it is not necessary, since in the C example, the wait loop in the end of `main` is slightly different from the C++ version.
- In the C++ version, before calling `.wait()` on every write image task, we must be sure that every element of the `image_write_tasks` contains a valid task item. This is ensured by calling `read_queue.finish();`, that synchronizes with the execution of the `read_complete_callback`s that create the entries in `image_write_tasks`.
- Whereas in the C version, we wait for `g_state.image_save_finished_count` to reach a predetermined value. Thereby it won't hurt if we enter the critical section *before* all the images were saved.

samples/core/callback/main.cpp
mfep: That's right. I didn't realize that `clEnqueueReadBuffer` holds ownership of the buffer. Fixed.

samples/core/callback/main.cpp
mfep: Agreed. Fixed.

samples/core/callback/main.cpp
mfep: Done as requested.

samples/core/callback/main.c
bashbaug: Oof, what a pain.  This is why we can't have nice things :smile: 

Thanks for the explanation, I'll just resolve this conversation.

samples/core/callback/CMakeLists.txt
bashbaug: Oof, what a pain.  This is why we can't have nice things :smile: 

Thanks for the explanation, I'll just resolve this conversation.

samples/core/callback/CMakeLists.txt
bashbaug: Weird, for my own reference (or in case somebody else stumbles across this):

https://stackoverflow.com/questions/33697072/why-is-pthread-necessary-for-usage-of-stdthread-in-gcc-and-clang

samples/core/callback/main.c
bashbaug: I actually only noticed this because I ran these samples with debugging and profiling tools enabled, specifically the [OpenCL Intercept Layer](https://github.com/intel/opencl-intercept-layer/blob/main/docs/chrome_tracing.md) and I didn't see any activity on the device in the C version.  This is because the C sample currently does not make any blocking calls that would otherwise cause the device activity to be "flushed".  I don't think this is necessarily a problem in the sample, but it's definitely odd and not something I've seen before!  I'll probably need to add provisions to handle this in the OpenCL Intercept Layer.

My one thought though is that without any blocking calls, calls to `clFinish`, or even calls to `clFlush`, if an implementation is maximally "lazy" and buffers everything, is there anything that would cause execution to begin?  I haven't seen this in any of the implementations that I've tested but I think it's possible in theory.

samples/core/callback/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+    target_link_libraries(callback PRIVATE
+        $<TARGET_NAME_IF_EXISTS:Threads::Threads>)
+else()
+    message(WARNING
+    "Skipping callback sample, C11 standard threads are not supported with the current toolset")
+endif()
+
+add_sample(
+    TEST
+    TARGET callback_cpp
```
bashbaug, line 48: Minor: the other SDK samples don't have an underscore before "cpp".  Should this be:

```suggestion
    TARGET callbackcpp
```

bashbaug (commented):
Thanks, I think this is getting close.  Please take a look at the one remaining comment above, and the naming comment below, otherwise all of my questions have been addressed.

samples/core/callback/main.c
mfep: Thank you for mentioning this tool! I didn't know it before, but it looks to be very useful. I could reproduce the same thing: device timing was only available when a call to `clFinish` was made. `clFlush` was not enough.

For know, I've updated the C example with a `clFlush` to all 3 queues. Based on what you suggest and the API spec, this seems to be the correct thing to do. Let me know if you find this a satisfying resolution.

samples/core/callback/CMakeLists.txt
mfep: Done as requested.

samples/core/callback/main.c
bashbaug: Thanks! I think the calls to `clFlush` should be sufficient.

I also just pushed a fix to the tool for this so I think we should be good.

https://github.com/intel/opencl-intercept-layer/pull/326

bashbaug (approved):

bashbaug:
Merging as discussed in the December 5th teleconference.

PR #90, opened 2023-10-27
Implemented external memory sample

mfep:
**NOTE**: this PR includes all commits from branch [StreamHPC:release-cd](https://github.com/StreamHPC/OpenCL-SDK/tree/release-cd). This is required, because the implemented sample requires additional updates to the CI scripts to pull its new dependencies. After #88 is merged, this should be trivial to rebase/merge. Only the last commit is relevant for the sample, all other commits are from the `release-cd` branch.

- Implemented sample showcasing the interop of OpenCL and Vulkan, using external memory,
- for details, please visit the newly added README file.

### DIFF
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -41,7 +41,7 @@ set(BUILD_SHARED_LIBS OFF CACHE BOOL "Global flag to cause add_library() to crea
 
 # Fetch dependencies
 if(OPENCL_SDK_BUILD_SAMPLES)
-  foreach(DEP IN ITEMS cargs TCLAP Stb Vulkan)
+  foreach(DEP IN ITEMS cargs TCLAP Stb)
     list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/Dependencies/${DEP}")
     include(${DEP})
   endforeach()
@@ -52,6 +52,11 @@ if(OPENCL_SDK_BUILD_SAMPLES)
       include(${DEP})
     endforeach()
   endif(OPENCL_SDK_BUILD_OPENGL_SAMPLES)
+
+  if(OPENCL_SDK_BUILD_VULKAN_SAMPLES)
+    list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/Dependencies/Vulkan")
+    include(Vulkan)
+  endif(OPENCL_SDK_BUILD_VULKAN_SAMPLES)
 endif(OPENCL_SDK_BUILD_SAMPLES)
 
 if(OPENCL_SDK_BUILD_CLINFO)
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -41,7 +41,7 @@ set(BUILD_SHARED_LIBS OFF CACHE BOOL "Global flag to cause add_library() to crea
 
 # Fetch dependencies
 if(OPENCL_SDK_BUILD_SAMPLES)
-  foreach(DEP IN ITEMS cargs TCLAP Stb)
+  foreach(DEP IN ITEMS cargs TCLAP Stb Vulkan)
     list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/Dependencies/${DEP}")
     include(${DEP})
   endforeach()
--- a/samples/extensions/khr/externalmemory/CMakeLists.txt
+++ b/samples/extensions/khr/externalmemory/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (c) 2020 The Khronos Group Inc.
+# Copyright (c) 2021 The Khronos Group Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -12,9 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-add_subdirectory(externalmemory)
-add_subdirectory(histogram)
-if(OPENCL_SDK_BUILD_OPENGL_SAMPLES)
-  add_subdirectory(conway)
-  add_subdirectory(nbody)
-endif()
+add_sample(
+    TEST
+    TARGET externalmemory
+    VERSION 300
+    SOURCES main.c
+    KERNELS external_saxpy.cl
+    LIBS Vulkan::Vulkan
+    DEFINITIONS $<$<PLATFORM_ID:Windows>:VK_USE_PLATFORM_WIN32_KHR>)
+
+add_sample(
+    TEST
+    TARGET externalmemorycpp
+    VERSION 300
+    SOURCES main.cpp
+    KERNELS external_saxpy.cl
+    LIBS Vulkan::Vulkan
+    DEFINITIONS $<$<PLATFORM_ID:Windows>:VK_USE_PLATFORM_WIN32_KHR>)
--- a/samples/extensions/khr/externalmemory/main.c
+++ b/samples/extensions/khr/externalmemory/main.c
@@ -551,9 +551,9 @@ int main(int argc, char* argv[])
 #else
         (cl_mem_properties)fd_x,
 #endif
-        (cl_mem_properties)CL_DEVICE_HANDLE_LIST_KHR,
+        (cl_mem_properties)CL_MEM_DEVICE_HANDLE_LIST_KHR,
         (cl_mem_properties)(uintptr_t)cl_device,
-        CL_DEVICE_HANDLE_LIST_END_KHR,
+        CL_MEM_DEVICE_HANDLE_LIST_END_KHR,
         0
     };
     cl_mem_properties ext_mem_props_y[] = {
@@ -563,9 +563,9 @@ int main(int argc, char* argv[])
 #else
         (cl_mem_properties)fd_y,
 #endif
-        (cl_mem_properties)CL_DEVICE_HANDLE_LIST_KHR,
+        (cl_mem_properties)CL_MEM_DEVICE_HANDLE_LIST_KHR,
         (cl_mem_properties)(uintptr_t)cl_device,
-        CL_DEVICE_HANDLE_LIST_END_KHR,
+        CL_MEM_DEVICE_HANDLE_LIST_END_KHR,
         0
     };
     cl_mem cl_buf_x, cl_buf_y;
--- a/samples/extensions/khr/externalmemory/main.c
+++ b/samples/extensions/khr/externalmemory/main.c
@@ -671,8 +671,7 @@ int main(int argc, char* argv[])
         printf("Kernel execution time as seen by host: %llu us.\n",
                (unsigned long long)(dev_time + 500) / 1000);
 
-        printf("Kernel execution time as measured by device:\n");
-        printf("\t%llu us.\n",
+        printf("Kernel execution time as measured by device: %llu us.\n",
                (unsigned long long)(cl_util_get_event_duration(
                                         kernel_run, CL_PROFILING_COMMAND_START,
                                         CL_PROFILING_COMMAND_END, &error)
--- a/samples/extensions/khr/externalmemory/main.c
+++ b/samples/extensions/khr/externalmemory/main.c
@@ -430,7 +430,11 @@ int main(int argc, char* argv[])
         goto arry;
     }
 
+    VkBufferUsageFlags vk_external_memory_usage =
+        VK_BUFFER_USAGE_TRANSFER_SRC_BIT | VK_BUFFER_USAGE_TRANSFER_DST_BIT;
+
     if (!vk_check_external_memory_handle_type(vk_physical_device,
+                                              vk_external_memory_usage,
                                               vk_external_memory_handle_type))
     {
         fprintf(stderr,
@@ -455,8 +459,7 @@ int main(int argc, char* argv[])
     buffer_info.sType = VK_STRUCTURE_TYPE_BUFFER_CREATE_INFO;
     buffer_info.pNext = &external_memory_buffer_info;
     buffer_info.size = sizeof(cl_float) * length;
-    buffer_info.usage =
-        VK_BUFFER_USAGE_TRANSFER_SRC_BIT | VK_BUFFER_USAGE_TRANSFER_DST_BIT;
+    buffer_info.usage = vk_external_memory_usage;
     buffer_info.sharingMode = VK_SHARING_MODE_EXCLUSIVE;
 
     VkBuffer vk_buf_x, vk_buf_y;
--- a/samples/extensions/khr/externalmemory/main.c
+++ b/samples/extensions/khr/externalmemory/main.c
@@ -354,36 +354,10 @@ int main(int argc, char* argv[])
                      context, 1, (const char**)&kernel, &program_size, &error),
                  error, ker);
 
-    // The Khronos extension showcased requires OpenCL 3.0 version.
-    // Get number of versions supported.
-    size_t versions_size = 0;
-    OCLERROR_RET(clGetDeviceInfo(cl_device, CL_DEVICE_OPENCL_C_ALL_VERSIONS, 0,
-                                 NULL, &versions_size),
-                 error, prg);
-    size_t versions_count = versions_size / sizeof(cl_name_version);
-
-    // Get and check versions.
-    cl_name_version* dev_versions = (cl_name_version*)malloc(versions_size);
-    OCLERROR_RET(clGetDeviceInfo(cl_device, CL_DEVICE_OPENCL_C_ALL_VERSIONS,
-                                 versions_size, dev_versions, NULL),
-                 error, prg);
-    char compiler_options[1024] = "";
-    for (cl_uint i = 0; i < versions_count; ++i)
-    {
-        if (opencl_version_is_major(&dev_versions[i], 3))
-        {
-            strcat(compiler_options, "-cl-std=CL3.0 ");
-        }
-    }
-
-    if (compiler_options[0] == '\0')
-    {
-        fprintf(stderr, "\nError: OpenCL version must be at least 3.0\n");
-        exit(EXIT_FAILURE);
-    }
-
-    OCLERROR_RET(cl_util_build_program(program, cl_device, compiler_options),
-                 error, prg);
+    // Build OpenCL executable.
+    OCLERROR_RET(
+        cl_util_build_program(program, cl_device, NULL /*compiler_options*/),
+        error, prg);
 
     // Query maximum workgroup size (WGS) supported based on private mem
     // (registers) constraints.
--- a/samples/extensions/khr/externalmemory/main.c
+++ b/samples/extensions/khr/externalmemory/main.c
@@ -430,7 +430,6 @@ int main(int argc, char* argv[])
     buffer_info.size = sizeof(cl_float) * length;
     buffer_info.usage =
         VK_BUFFER_USAGE_TRANSFER_SRC_BIT | VK_BUFFER_USAGE_TRANSFER_DST_BIT;
-    ;
     buffer_info.sharingMode = VK_SHARING_MODE_EXCLUSIVE;
 
     VkBuffer vk_buf_x, vk_buf_y;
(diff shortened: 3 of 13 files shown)


nikhiljnv (approved):
Large change to review meaningfully:)
Approving based on our successful testing on NVIDIA implementation.

cmake/Dependencies.cmake
```diff
@@ -34,7 +34,7 @@ set(BUILD_SHARED_LIBS OFF CACHE BOOL "Global flag to cause add_library() to crea
 
 # Fetch dependencies
 if(OPENCL_SDK_BUILD_SAMPLES)
-  foreach(DEP IN ITEMS cargs TCLAP Stb)
+  foreach(DEP IN ITEMS cargs TCLAP Stb Vulkan)
```
bashbaug, line 37: I think this makes Vulkan a dependency for all of the samples, correct?  It might be best to add an `OPENCL_SDK_BUILD_VULKAN_SAMPLES` option similar to `OPENCL_SDK_BUILD_OPENGL_SAMPLES`, below, so the non-Vulkan samples will continue to work even on platforms where Vulkan may be unavailable.

samples/extensions/khr/externalmemory/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+# limitations under the License.
+
+add_sample(
+    TEST
+    TARGET externalmemory
+    VERSION 300
+    SOURCES main.c
+    KERNELS external_saxpy.cl
+    LIBS Vulkan::Vulkan
+    DEFINITIONS $<$<PLATFORM_ID:Windows>:VK_USE_PLATFORM_WIN32_KHR>)
```
bashbaug, line 22: This is somewhat personal preference, but I find it simpler to add these types of conditional defines in the program source itself vs. CMake logic.

```c
#ifdef _WIN32
#define VK_USE_PLATFORM_WIN32_KHR
#endif
```

samples/extensions/khr/externalmemory/main.cpp
```diff
@@ -0,0 +1,559 @@
...
+        fd_info_y.pNext = nullptr;
+        fd_info_y.memory = vk_buf_y_memory;
+        fd_info_y.handleType = vk_external_memory_handle_type;
+        int fd_y;
+
+        // We need to get the pointer to the vkGetMemoryFdKHR function because
+        // it's from extension VK_KHR_external_memory_fd.
+        PFN_vkGetMemoryFdKHR vkGetMemoryFdKHR =
+            (PFN_vkGetMemoryFdKHR)vkGetDeviceProcAddr(vk_device,
+                                                      "vkGetMemoryFdKHR");
```
bashbaug, line 427: This is the function that is needed for Linux and file descriptors, but a different function is needed for Windows and NT handles.  Has this sample been tested on Windows?

samples/extensions/khr/externalmemory/vulkan_utils.h
```diff
@@ -0,0 +1,382 @@
...
+        (cl_platform_id*)malloc(cl_platform_count * sizeof(cl_platform_id));
+    OCLERROR_RET(clGetPlatformIDs(cl_platform_count, platforms, NULL), error,
+                 platforms);
+
+    size_t cl_device_count = 0;
+    const char* uuid_khronos_extension[] = {
+        CL_KHR_DEVICE_UUID_EXTENSION_NAME
+    };
+    for (cl_uint cl_platform_id = 0; cl_platform_id < cl_platform_count;
+         ++cl_platform_id)
```
bashbaug, line 241: It'd be best not to name this loop variable `cl_platform_id` since that is also the name of an OpenCL type.

samples/extensions/khr/externalmemory/vulkan_utils.h
```diff
@@ -0,0 +1,382 @@
...
+                memcpy(candidate.uuid, &vk_candidate_uuid,
+                       sizeof(cl_uchar) * CL_UUID_SIZE_KHR);
+                cl_candidates[cl_device_count] = candidate;
+            }
+        }
+    }
+
+    // Query the Vulkan physical devices available.
+    uint32_t vk_device_count;
+    VK_CHECK(vkEnumeratePhysicalDevices(instance, &vk_device_count, NULL));
```
bashbaug, line 301: FWIW the C version of the sample crashes here on my Linux system.  I haven't debugged further.

```gdb
Thread 1 "externalmemory" received signal SIGSEGV, Segmentation fault.
0x00007ffff7ca171f in unlink_chunk (p=p@entry=0x555555723dc0, av=0x7ffff7e19c80 <main_arena>) at ./malloc/malloc.c:1628
1628	./malloc/malloc.c: No such file or directory.
(gdb) bt
#0  0x00007ffff7ca171f in unlink_chunk (p=p@entry=0x555555723dc0, av=0x7ffff7e19c80 <main_arena>) at ./malloc/malloc.c:1628
#1  0x00007ffff7ca462b in _int_malloc (av=av@entry=0x7ffff7e19c80 <main_arena>, bytes=bytes@entry=64) at ./malloc/malloc.c:4307
#2  0x00007ffff7ca5139 in __GI___libc_malloc (bytes=64) at ./malloc/malloc.c:3329
#3  0x00007fffe75b7f65 in ?? () from /usr/lib/x86_64-linux-gnu/libvulkan_intel.so
#4  0x00007fffe75b812b in ?? () from /usr/lib/x86_64-linux-gnu/libvulkan_intel.so
#5  0x00007fffe75b9643 in ?? () from /usr/lib/x86_64-linux-gnu/libvulkan_intel.so
#6  0x00007fffe74fe397 in ?? () from /usr/lib/x86_64-linux-gnu/libvulkan_intel.so
#7  0x00007fffe74fe982 in ?? () from /usr/lib/x86_64-linux-gnu/libvulkan_intel.so
#8  0x00007fffe748eea9 in ?? () from /usr/lib/x86_64-linux-gnu/libvulkan_intel.so
#9  0x00007fffe729cd3c in ?? () from /usr/lib/x86_64-linux-gnu/libvulkan_intel.so
#10 0x00007fffe747c3c1 in ?? () from /usr/lib/x86_64-linux-gnu/libvulkan_intel.so
#11 0x00007fffe747cef1 in ?? () from /usr/lib/x86_64-linux-gnu/libvulkan_intel.so
#12 0x00007ffff7f49718 in ?? () from /lib/x86_64-linux-gnu/libvulkan.so.1
#13 0x00007ffff7f49dec in ?? () from /lib/x86_64-linux-gnu/libvulkan.so.1
#14 0x00007fffed99b40c in ?? () from /lib/x86_64-linux-gnu/libVkLayer_MESA_device_select.so
#15 0x00007ffff7f53e41 in vkEnumeratePhysicalDevices () from /lib/x86_64-linux-gnu/libvulkan.so.1
#16 0x000055555555736d in find_suitable_device (instance=0x5555555623d0, required_device_extensions=0x555555561050 <required_device_extensions>, required_device_extensions_count=2)
    at /home/bashbaug/git/opencl-sdk/samples/extensions/khr/externalmemory/vulkan_utils.h:301
#17 0x0000555555557d76 in main (argc=1, argv=0x7fffffffe008) at /home/bashbaug/git/opencl-sdk/samples/extensions/khr/externalmemory/main.c:264
(gdb) 
```

bashbaug (commented):
Note: I only reviewed the last commit.

At some point it would be good to update this sample so it worked with other types of external memory as well, e.g. dma_buf file descriptors.

samples/extensions/khr/externalmemory/main.cpp
```diff
@@ -0,0 +1,559 @@
...
+            std::cout << "done.\nExecuting on device... ";
+            std::cout.flush();
+        }
+
+        std::vector<cl::Event> kernel_run;
+        auto dev_start = std::chrono::high_resolution_clock::now();
+        kernel_run.push_back(
+            saxpy(cl::EnqueueArgs{ queue, cl::NDRange{ length }, wgs }, a,
+                  cl_buf_x, cl_buf_y));
+        cl::WaitForEvents(kernel_run);
```
bashbaug, line 472: One more - this is missing the calls to acquire and release the external memory objects.

See: https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#_acquiring_and_releasing_external_memory_objects

bashbaug (changes requested):
Note: I've left comments on the C++ sample, but many of the issues occur in the C sample also.  Thanks!

cmake/Dependencies.cmake
Beanavil: Added!

samples/extensions/khr/externalmemory/vulkan_utils.h
Beanavil: Changed!

samples/extensions/khr/externalmemory/vulkan_utils.h
Beanavil: Hmm for the Linux environments tested it works fine ([this](https://github.com/StreamHPC/OpenCL-SDK/blob/external-mem-sample/.github/workflows/presubmit.yml#L31-L126) is the matrix of combinations tested), could you provide more info about your setup to try to reproduce this?

samples/extensions/khr/externalmemory/main.cpp
Beanavil: Added!

samples/extensions/khr/externalmemory/main.cpp
Beanavil: Yes, this sample has been tested on Windows, but we got 
```
Test #31: externalmemory_Release ...........Exit code 0xc0000135
***Exception:   0.01 sec

Test #32: externalmemorycpp_Release ........Exit code 0xc0000135
***Exception:   0.01 sec
```
In the latest commit, I have added the logic to get the function pointer and call the correct function on Windows, but the runtime fail is still there: https://github.com/StreamHPC/OpenCL-SDK/actions/runs/7216235279/job/19662376992. This runtime error also appears on MacOS, although for now both tests are disabled on CI ([macos' here](https://github.com/StreamHPC/OpenCL-SDK/blob/external-mem-sample/.github/workflows/presubmit.yml#L743-L745) and [windows' here](https://github.com/StreamHPC/OpenCL-SDK/blob/external-mem-sample/.github/workflows/presubmit.yml#L571-L578)).

Given that it works fine in Linux, we assumed this failure could be due to no appropriate Vulkan device present on the failing test machines from CI. Perhaps you have more insights about this?

samples/extensions/khr/externalmemory/CMakeLists.txt
Beanavil: If it's not a problem, we prefer to keep this logic :)

samples/extensions/khr/externalmemory/main.cpp
bashbaug: I think this is still failing because the handle type needs to be updated for Windows:

```c++
            (cl_mem_properties)CL_EXTERNAL_MEMORY_HANDLE_OPAQUE_FD_KHR,
#ifdef _WIN32
            (cl_mem_properties)handle_y,
#else
            (cl_mem_properties)fd_y,
#endif
```

samples/extensions/khr/externalmemory/vulkan_utils.h
bashbaug: I'm still getting this crash on my system with the latest code.

> could you provide more info about your setup to try to reproduce this?

It's a pretty standard Ubuntu 22.04 system, nothing too fancy.

The main thing that is a little different is that it has several GPUs in it, including two discrete GPUs and one integrated GPU.  I can't think of a reason why that might cause this failure, though, and only with the C version of the sample?  The C++ version is working fine.

bashbaug: I think I found the problem.

The issue is that one of my OpenCL devices (POCL) does not support the `cl_khr_device_uuid` extension, so it does not increment `cl_device_count` when determining the number of candidate devices (the initial loop in `find_suitable_device`).  When building the list of candidates (the second loop in `find_suitable_device`) the `cl_device_count` gets incremented unconditionally, though.  This means instead of skipping the POCL device entirely, there is a "hole" in the candidate device list, and we end up writing more candidate devices than we've allocated.

The fix is either to increment the `cl_device_count` conditionally when building the candidate device list, consistent with the way the number of candidates was computed.

bashbaug:
There is a merge conflict here that needs to be resolved, also.

samples/extensions/khr/externalmemory/vulkan_utils.h
Beanavil: Certainly, the logic was faulty as you describe. I added a commit fixing this, so that:
1. We do not continue checking the Vulkan-compatible devices if there were no OpenCL-compatible devices to begin with (`cl_device_count == 0`)
2. While checking the Vulkan-compatible devices we only increment the `cl_device_count` if the device supports the required extension

Beanavil:
Addressed the comments and rebased onto StreamHPC/release-cd. There are some test failures from the other tests of the repo, but I think those are due to `release-cd` still being worked on to solve some new issues after rebasing. It's probably better to wait until #88  is ready to merge and rebase this one onto it.

samples/extensions/khr/externalmemory/main.cpp
bashbaug: I think this still needs to be fixed.  Have these samples been tested on Windows?

samples/extensions/khr/externalmemory/main.c
```diff
@@ -0,0 +1,714 @@
...
+        printf("Verification passed.\n");
+    }
+
+    if (!diag_opts.quiet)
+    {
+        printf("Kernel execution time as seen by host: %llu us.\n",
+               (unsigned long long)(dev_time + 500) / 1000);
+
+        printf("Kernel execution time as measured by device:\n");
+        printf("\t%llu us.\n",
```
bashbaug, line 675: Super minor: the CPP sample doesn't include an extra newline here, and prints the device time on the same line, similar to the host time.

bashbaug (commented):
The C sample is fixed and doesn't crash for me anymore, thanks!

I think the Windows sample is still broken, though.

samples/extensions/khr/externalmemory/main.cpp
Beanavil: Forgot to answer this, sorry. The handle type is allegedly fixed, but the example still fails on Windows. Right now is disabled on CI, but I tested it on my personal fork https://github.com/Beanavil/OpenCL-SDK/actions/runs/9563524333/job/26362215011.

samples/extensions/khr/externalmemory/main.c
Beanavil: Changed as requested!

samples/extensions/khr/externalmemory/main.cpp
```diff
@@ -0,0 +1,635 @@
...
+            cl_context,
+            std::string{ std::istreambuf_iterator<char>{ kernel_stream },
+                         std::istreambuf_iterator<char>{} }
+        };
+
+        // The Khronos extension showcased requires OpenCL 3.0 version.
+        cl::string compiler_options = "";
+        const std::string dev_version =
+            cl_device.getInfo<CL_DEVICE_OPENCL_C_VERSION>();
+        if (opencl_version_contains(dev_version, "3."))
```
bashbaug, line 278: This isn't the right way to check for this.  [CL_DEVICE_OPENCL_C_VERSION](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_OPENCL_C_VERSION) is the older query that returns the "highest fully backwards compatible OpenCL C version supported by the compiler for the device".  This means that many OpenCL 3.0 devices will only return "OpenCL 1.2" for this query.

The right way to query for OpenCL C 3.0 support is via [CL_DEVICE_OPENCL_C_ALL_VERSIONS](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_OPENCL_C_ALL_VERSIONS).

This really ought to be wrapped in an SDK utility function.

samples/extensions/khr/externalmemory/vulkan_utils.h
```diff
@@ -0,0 +1,389 @@
...
+        CL_KHR_DEVICE_UUID_EXTENSION_NAME
+    };
+    for (cl_uint platform_id = 0; platform_id < platform_count;
+         ++platform_id)
+    {
+        cl_uint cl_platform_devices_count = 0;
+        OCLERROR_RET(clGetDeviceIDs(platforms[platform_id],
+                                    CL_DEVICE_TYPE_ALL, 0, NULL,
+                                    &cl_platform_devices_count),
+                     error, platforms);
```
bashbaug, line 247: It would be best not to check for an error here, or at least to filter out the `CL_DEVICE_NOT_FOUND` error.  If a platform supports zero devices (maybe the device is powered down or otherwise disabled) then this should not prevent the sample from executing on a different device.

FYI: https://github.com/KhronosGroup/OpenCL-Docs/issues/1269

bashbaug (changes requested):
FYI, here are a few more issues I found.

Note that the sample still does not execute properly on Windows.

samples/extensions/khr/externalmemory/main.cpp
Beanavil: Arg yes you're right, fixed this now.

Btw, I checked whether this was used in other examples and found out it was also used in the [reduce](https://github.com/KhronosGroup/OpenCL-SDK/blob/main/samples/core/reduce/main.c#L148) core sample. I think it should be changed there as well.

samples/extensions/khr/externalmemory/vulkan_utils.h
Beanavil: Changed this! The Platform::getDevices functionality from the C++ headers (which is used in the vulkan_utils.cpp) also follows the logic you described, so this change also fixes consistency between the C and C++ versions of the sample

samples/extensions/khr/externalmemory/vulkan_utils.h
```diff
@@ -0,0 +1,402 @@
...
+#include <stdlib.h>
+#include <string.h>
+
+// Check if
+#define CL_CHECK_NO_DEVICE(func, err, label)                                   \
+    do                                                                         \
+    {                                                                          \
+        err = func;                                                            \
+        if (err != CL_SUCCESS) goto label;                                     \
+    } while (0)
```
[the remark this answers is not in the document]
Beanavil: Done!

Beanavil:
@bashbaug about the Windows crash: unfortunately, I'm only able to run this sample in Windows through the CI and debugging it is rather slow. I checked the error that the sample produced, and it was the `0xc0000135`, which means there was a missing DLL (the `vulkan-1.dll` one it seems). As expected, installing the Vulkan loader fixed that error, but then the example still fails with a `0xc0000139`. I think this error is an "Entry point not found" (some DLL is being loaded but that cannot find the other DLLs it relies on), so it may be that other Vulkan-related dependencies are missing. I'll look further into this sometime this week if that's okay.

Beanavil:
- Rebased the branch onto main because of merge conflicts
- Added the fixes discussed in https://github.com/StreamHPC/OpenCL-SDK/pull/8

samples/extensions/khr/externalmemory/main.c
```diff
@@ -0,0 +1,746 @@
...
+        error, err);
+    supported_handle_types = (cl_external_memory_handle_type_khr*)malloc(
+        supported_handle_types_count);
+
+    OCLERROR_RET(
+        clGetDeviceInfo(
+            cl_device, CL_DEVICE_EXTERNAL_MEMORY_IMPORT_HANDLE_TYPES_KHR,
+            supported_handle_types_count, supported_handle_types, NULL),
+        error, err);
+    for (size_t i = 0; i < supported_handle_types_count; ++i)
```
bashbaug, line 202: `supported_handle_types_count` is actually the size in bytes rather than a count, so this loop will read off the end of the allocation.  There needs to be a divide by `cl_external_memory_handle_type_khr` somewhere to compute the count.

samples/extensions/khr/externalmemory/main.c
```diff
@@ -0,0 +1,746 @@
...
+    OCLERROR_RET(clGetDeviceInfo(cl_device, CL_DEVICE_OPENCL_C_ALL_VERSIONS,
+                                 versions_size, dev_versions, NULL),
+                 error, prg);
+    char compiler_options[1024] = "";
+    for (cl_uint i = 0; i < versions_count; ++i)
+    {
+        if (opencl_version_is_major(&dev_versions[i], 3))
+        {
+            strcat(compiler_options, "-cl-std=CL3.0 ");
+        }
```
bashbaug, line 372: I don't see anything in the `external_saxpy.cl` kernel that requires OpenCL C 3.0.  Am I missing something?  If not, we could remove this code block to shorten the sample, and it could run on more devices also (in theory, at least).

bashbaug: Note, if we remove this code block from the C sample, we should remove it from the C++ sample also.

bashbaug (approved):
I'd like to fix the first issue below if possible, but otherwise I think this is ready to go.  If we find any additional issues we can fix them after merging.

samples/extensions/khr/externalmemory/main.c
Beanavil: Removed! OpenCL 3.0 is certainly only necessary for the external memory extension.

PR #99, opened 2024-01-26
remove unused symlinks

### LINKED ISSUE #82: Remove the symlink in the releases sources
On Windows, without activate the developer mode it cannot extract an archive that have symlink inside. So I propose to modify the release workflow to move the content of submodule in place of the symlink.

image

bashbaug:
These symlinks cause issues when the SDK is deployed on some systems.

Fixes #82

### DIFF
--- a/include/api/CL
+++ b/include/api/CL
@@ -1,1 +1,0 @@
-../../external/OpenCL-Headers/CL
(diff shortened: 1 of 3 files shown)


MathiasMagnus (approved):
As discussed on the WG meeting, these folders are remnants of an earlier design that serve no purpose in the current submodule-driven implementation.

PR #103, opened 2024-04-26
update project version and submodules - April 2024

bashbaug:
Update project version and submodules for an upcoming release.


bashbaug:
I've temporarily disabled macos for the Python and Ruby tests since something seems to have changed in the GitHub actions machine config, and now CI is clean.  We need to investigate what happened to cause macos to fail.  We can either do that before this PR is merged, or I will file an issue to investigate it after the PR is merged.

MathiasMagnus (approved):

PR #108, opened 2024-07-15
throw() -> noexcept

r-barnes:
`throw()` was deprecated in C++11 and removed in C++17. `noexcept` is the appropriate replacement.


bashbaug (approved):
No reason to wait on this one IMHO.

At some point we should clean up the documentation in Utils.md similarly.

Issue #112, opened 2024-10-01
confirm intended install directories after merging CI/CD changes


bashbaug:
We'll check that the install directories are working as we intend after merging.  Specifically, with a non-multi-config generator (e.g. Makefiles), does the install go into per-config directories, or into a single `lib` or `bin` directory?

_Originally posted by @bashbaug in https://github.com/KhronosGroup/OpenCL-SDK/issues/88#issuecomment-2386424254_

bashbaug:
I checked the install directories on my Linux system with the latest changes, after merging #88.  Here is what I found, after building with the Makefile generator:

1. The OpenCL headers and C++ bindings go into a single "include" directory.
2. The OpenCL ICD loader goes into a single "lib" directory.
3. The cllayerinfo utility that is built with the OpenCL ICD loader goes into a single "bin" directory.
4. clinfo also goes into the same single "bin" directory.
5. The OpenCL SDK utility headers go into a single "include" directory.
6. The OpenCL SDK libraries go into a single "lib" directory.
7. Any supporting libraries for the samples (e.g. sfml, cargs) goes into the same single "lib" directory.
8. The sample applications and kernels go into a per-config "bin" directory, for example "bin/Release".

This seems like the opposite of what was described in the October 1st teleconference.

Do we intend for the sample applications and kernels in (8) to be installed into a per-config "bin" directory?  Note, this is different behavior than we had before merging #88.

CC @MathiasMagnus @Beanavil

Beanavil:
@bashbaug hm yes looking again at the logic of this CMakeLists file it does seem like it will install the kernels and executables in `/<install_dir>/bin/<config>` regardless of the generator used. If we want to maintain the previous behaviour for the non-multiconfig generators, I think this is easily fixable.

However, I've been thinking about why would this be a problem. You did mention that it could affect packaging, but as far as I can remember there was no binary packaging for the SDK before and also afaik we do not package the samples' binaries anyway, or am I missing something here?

MathiasMagnus:
@bashbaug Generally speaking, the OpenCL-SDK's build tree isn't guaranteed to be stable, not that it's even possible, given how different it is with auto-built deps vs. user-provided pre-built deps. Our install tree is intended to be stable and wishes to emulate a proper *nix layout, mostly replicated by the [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) module. As such the intention was/is:

[...]

PR #113, opened 2024-10-10
Revert per-config samples install directories

Beanavil:
This PR reverts the changes made in #88 in which the install directories of the samples were modified to per-config install directories inside the `bin` folder.

Additionally, the MacOS CI job has been patched as it seems that the install directory of PoCL has changed (from `/opt/homebrew/Cellar/pocl/6.0` to `/opt/homebrew/Cellar/pocl/6.0_1`) and thus the hard-coded vendors path used within the workflow was outdated (and triggering tests failures).


Beanavil:
@bashbaug fyi: agreed with @MathiasMagnus to split the TODO work for SDK and Layers

bashbaug (approved):
LGTM, thanks!

Does this fix #112 or is additional work needed?

Beanavil:
@bashbaug  No further work is needed AFAIK

Issue #121, opened 2024-12-10
Where is "include" folder in repo?


dmmil:
I can't run the example from "OpenCL-SDK/samples/core/binaries" because header files are missing. The project description says that "It brings together all the components needed to develop OpenCL applications:
OpenCL Headers (include/api)
OpenCL C++ bindings (include/cpp)
OpenCL Loader
OpenCL utility library (include/utils)",
however, there is no include folder in the repository at all. Please tell me where to get these files?

bashbaug:
Hi, this project uses sub-modules for the OpenCL external dependencies.  Can you please follow the instructions here?

https://github.com/KhronosGroup/OpenCL-SDK?tab=readme-ov-file#dependencies

I see though that the README has the incorrect paths to these sub-modules.  We should fix that.  I believe the correct paths are:

- OpenCL Headers (`external/OpenCL-Headers/`)
- OpenCL C++ bindings (`external/OpenCL-CLHPP/include`)
- OpenCL Loader (`external/OpenCL-ICD-Loader`)
- OpenCL utility library (`lib/include`)

PR #122, opened 2024-12-10
update the paths to OpenCL include files and libraries in the README

bashbaug:
found while responding to #121 

The paths described in the README do not exist.  This PR fixes them.


MathiasMagnus (approved):
Yes, the submodule transition would've need these to be updated. Thanks for fixing.

PR #124, opened 2024-12-13
Add Ocean OpenCL-OpenGL interop demo

starcatter:
For comparison purposes with #110 

The sample follows general steps (with multiple optimizations) described in the publication: Realtime GPGPU FFT ocean water simulation

Main focus of the sample is to demonstrate how to share compute/render resources between OpenCL and OpenGL to simulate an ocean surface.


samples/extensions/khr/ocean/ocean_util.hpp
```diff
@@ -0,0 +1,194 @@
...
+            default: ErrorString = "UNKNOWN"; break;
+        }
+        std::cerr << "OpenGL Error(" << ErrorString << "): " << Title
+                  << std::endl;
+    }
+    return Error == GL_NO_ERROR;
+}
+
+
+static const char* IGetErrorString(int clErrorCode)
```
bashbaug, line 61: This seems generally useful.  Could we move this to a shared location?  Or, could we use the `cl_util_print_error` function already in the SDK utils lib?

samples/extensions/khr/ocean/main.cpp
```diff
@@ -0,0 +1,194 @@
...
+template <>
+CliOptions cl::sdk::comprehend<CliOptions>(
+    std::shared_ptr<TCLAP::ValueArg<bool>> useGLSharing)
+{
+    return CliOptions{ useGLSharing->getValue() };
+}
+
+int main(int argc, char* argv[])
+{
+    OceanApplication app;
```
bashbaug, line 153: Because `app` is instantiated before parsing command line options, it is ignoring any passed-in command line options to select the platform or device index.  Could we instead parse the command line options first and pass the chosen platform index, device index, and device type when we construct `app`?  See the nbody sample as an example.

bashbaug (commented):
This is great work and I think we should try to get it and #110 in for the next release.

This PR seems to share a lot of files (e.g. OpenCL kernels) with #110.  Should they be renamed so they are distinct?  Or, is the intent that both samples will share the same OpenCL kernels?

bashbaug:
Hi @starcatter, thanks again for your contribution.  Do you think you will have a chance to address the review comments above?  Thanks!

starcatter:
> This is great work and I think we should try to get it and #110 in for the next release.
> 
> This PR seems to share a lot of files (e.g. OpenCL kernels) with #110. Should they be renamed so they are distinct? Or, is the intent that both samples will share the same OpenCL kernels?

Yes, the intent was to use the same kernels in both samples. Only the shaders have subtle differences.

samples/extensions/khr/ocean/ocean_util.hpp
starcatter: Done.

samples/extensions/khr/ocean/main.cpp
starcatter: Done.

samples/extensions/khr/ocean/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+add_sample(
+    TARGET ocean_gl_ocl_interop
+    VERSION 300
+    SOURCES main.cpp ocean.cpp ocean.hpp ocean_util.hpp
+    KERNELS twiddle.cl time_spectrum.cl inversion.cl normals.cl fft_kernel.cl init_spectrum.cl
+    SHADERS ocean.vert.glsl ocean.frag.glsl)
```
bashbaug, line 20: I wasn't able to build on my system without adding a `glm` dependency also:

```suggestion
    SHADERS ocean.vert.glsl ocean.frag.glsl
    LIBS glm::glm)
```

bashbaug (commented):
Couple other things will need to be fixed before we can merge:

1. Can you please check that the CLA is all squared away?  It's possible some of the latest commits were made with the wrong GitHub ID.
2. Can you please take a look at the CI build errors?  I didn't see them on my system, but we'll need to fix them one way or the other before merging.

samples/extensions/khr/ocean/main.cpp
bashbaug: This still didn't work for me, at least for the platform and device arguments (specified with "-p"), and it looks like the sample is exclusively running on platform zero.  With my current config, this just so happens to be a CPU platform, so the sample isn't working.  I'd really like to be able to select the platform to run with.  Alternatively, we could try to find the right platform based on the OpenGL context with CL-GL sharing is enabled, but this isn't a viable solution when CL-GL sharing is disabled.

samples/extensions/khr/ocean/CMakeLists.txt
shajder: Applied, thanks.

samples/extensions/khr/ocean/main.cpp
shajder: I see similar behavior at my end -  the failure was a CL/GL context mismatch on a hybrid-GPU laptop, the OpenGL context was landing on the integrated GPU while OpenCL picked the discrete one. Forcing GL onto the same physical device via PRIME render offload, both -p and -d select the platform/device as expected. In other words, -p/-d select correctly; the platform and device you get were never the problem, only the fact that OpenGL was rendering on a different physical GPU than the one OpenCL was told to use.

shajder:
> 2\. Can you please take a look at the CI build errors?  I didn't see them on my system, but we'll need to fix them one way or the other before merging.

Fixed, CI is green across Linux, macOS, and Windows now.

PR #131, opened 2025-02-28
Do not build CL/GL samples by default

kpet:
These require many dependencies and it seems fetching the dependencies of SFML is currently broken. Better to disable them by default so first time users do not need to resolve complex dependency issues to get going.


bashbaug:
Discussed in the March 11th teleconference.

* @kpet will ensure the graphical samples are still tested in CI after this change.
* Additionally, need to update the README to describe how to enable the graphical samples.

bashbaug (approved):
Discussed in the April 1st teleconference.  OK to merge once we see passing CI.

kpet:
Merging as discussed in the 2025/04/08 teleconference.

PR #132, opened 2025-03-31
Sfml ver fix

MathiasMagnus:
All Vcpkg CI elements started failing recently, which is fixed by this PR. The core of the issue was that SFML, a dependency of the graphical samples built in CI have released a new major version, SFML3, which is incompatible with SFML2. It has slightly different CMake-side support, as well as slightly different interfaces. The fix involves pinning the version of SFML fetched from CMake, which can only be done via a manifest file; it's not possible via the CLI.

## Notes for the reviewer

- There are two commits in this PR. The first strictly deals with introducing the manifest file for Vcpkg and updating autofetching to the latest 2.x version available.
  - Caching Vcpkg installs got removed for the time being. Partly because it needs to be done differently (given how it's not a separate CLI invocation now, but part of the configuration process), but also because the keys used to identify the caches needed a facelift and it requires extra testing to make sure it works inside the configuration process to begin with. Vcpkg [has its own caching](https://learn.microsoft.com/en-us/vcpkg/users/binarycaching) mechanism too, so it may even prove to be unnecessary, or it can be reintroduced later with a proper key holding tool version strings.
  - Windows builds were using the system Vcpkg install un-updated, so were potentially installing different versions of deps than the Linux and MacOS builds.
- The second commit deals with a number of issues (skeletons in the closet). Solving the Vcpkg issue uncovered some unintended shortcomings.
  - Compiler flags were given to CMake via environmental variables, which unintentionally propagate into Vcpkg builds. Naturally, not all ports build warning-free with high warning levels.
  - Giving compiler flags was also not done properly, so 32-bit builds were actually 64-bit. Once fixed, DEB packaging turned out to be incompatible with 64-bit systems (like in CI). Proper multi-arch packaging requires some extra care apparently. (They should work fine on 32-bit only systems.)
  - 32-bit builds seem to require explicitly specifying the make program in order for Vcpkg to pick them up, therefore these were made explicit.
  - Getting proper 32-bit builds uncovered we need to install the 32-bit versions of the libs SFML depends on and needs from the system. (The CI image has the native 64-bit version installed)

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -1,0 +1,21 @@
+{
+	"$schema": "https://raw.githubusercontent.com/microsoft/vcpkg-tool/main/docs/vcpkg.schema.json",
+	"name": "opencl",
+	"builtin-baseline": "21816e0df975ab3ba13ab75263c4deeb080ef681",
+	"dependencies": [
+		{
+			"name": "sfml",
+			"version>=": "2.0"
+		},
+		"tclap",
+		"glm",
+		"glew",
+		"stb"
+	],
+	"overrides": [
+		{
+			"name": "sfml",
+			"version": "2.6.2"
+		}
+	]
+}
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -133,6 +133,7 @@ jobs:
       CMAKE_EXE: /opt/Kitware/CMake/${{ matrix.CMAKE }}/bin/cmake
       CPACK_EXE: /opt/Kitware/CMake/${{ matrix.CMAKE }}/bin/cpack
       CTEST_EXE: /opt/Kitware/CMake/${{ matrix.CMAKE }}/bin/ctest
+      VCPKG_INSTALLATION_ROOT: /opt/Microsoft/vcpkg
       DEB_INSTALLATION_PATH: /usr
       CC: ${{matrix.COMPILER.C_NAME}}-${{matrix.COMPILER.VER}}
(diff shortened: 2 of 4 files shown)


bashbaug (approved):
I don't use vcpkg often, so my approval shouldn't carry much weight, but this all looks reasonable to me and it looks like CI is passing again.

Do we need to update anything in the README after this change?  The [example build](https://github.com/KhronosGroup/OpenCL-SDK?tab=readme-ov-file#example-build) currently uses vcpkg.

vcpkg.json
```diff
@@ -0,0 +1,21 @@
...
+		"glew",
+		"stb"
+	],
+	"overrides": [
+		{
+			"name": "sfml",
+			"version": "2.6.2"
+		}
+	]
+}
```
Beanavil, line 21: nitpick (can be ignored):
```suggestion
}

```

Beanavil (approved):
lgtm!

MathiasMagnus:
@Beanavil @bashbaug I had to disable all the Vcpkg jobs, because the Vcpkg repo was [decimated](https://github.com/microsoft/vcpkg/issues/44726) by the 4.0 shift. It's outside our control and will likely get a fix, because a lot of people's workflows are broken now. (Vcpkg fetches a latest CMake for the sub-builds and doesn't use the parent process' executable.)

Please squash merge if CI passes and you agree with the final changes.

bashbaug (approved):
Could you please file an issue so we don't forget to re-enable the vcpkg testing once the CMake smoke clears?

Otherwise, LGTM.

PR #135, opened 2025-04-15
Update cargs dependency to v1.2.0

MichaelRizkalla-arm:
This patch updates cargs dependency to v1.2.0 and removes the custom command to patch CMakeLists.txt


bashbaug (approved):
Merging as discussed in the May 6th teleconference.

PR #136, opened 2025-07-04
Fix blur example usage of clGetDeviceInfo

MichaelRizkalla-arm:
`blur` example is retrieving `CL_DEVICE_OPENCL_C_VERSION` in a fixed sized array of 64 bytes, CL spec does not specify a limit to the output length. Therefore, the test may fail if the version is longer than 64 bytes.

This change fixes the issue by querying `clGetDeviceInfo` for the expected output size.


bashbaug:
Can't think of a reason to wait to merge this one, so.... merging!

PR #140, opened 2025-07-24
simplify the CMake build process for clinfo in the OpenCL SDK

bashbaug:
The way we have currently integrated clinfo into the OpenCL SDK is overly complicated, involving a custom modified file that replaces one of the clinfo files.  We can simplify our build process by using clinfo as-is, instead.


bashbaug:
Hmm, there is an issue with macOS that I am going to need to resolve.  Moving to "draft".

PR #143, opened 2025-08-07
Fix CL_INVALID_WORK_ITEM_SIZE for "reduce" test

xinjin01:
This PR fixes issues in the `reduce` sample where the chosen work-group size (WGS)
could exceed the device’s per-dimension maximum (`CL_DEVICE_MAX_WORK_ITEM_SIZES[0]`),
causing `clEnqueueNDRangeKernel` to fail with `CL_INVALID_WORK_GROUP_SIZE`.

The `reduce` sample has two implementations (C and C++), and both required fixes:

- **C version**: The first patch ensures the kernel is launched with a valid WGS by
  clamping against the device’s maximum.
- **C++ version**: The second patch applies the same clamping logic to the C++ `reduce.cpp`
  application, ensuring consistent behavior.

Together, these patches make both variants of the `reduce` sample more robust and
portable across a wider range of OpenCL devices.

Signed-off-by: Xin Jin <xin.jin@arm.com>

### DIFF
--- a/samples/core/reduce/main.c
+++ b/samples/core/reduce/main.c
@@ -386,6 +386,15 @@ int main(int argc, char *argv[])
     OCLERROR_RET(clGetDeviceInfo(device, CL_DEVICE_LOCAL_MEM_SIZE,
                                  sizeof(cl_ulong), &loc_mem, NULL),
                  error, red);
+
+    // Ensure WGS is valid for this device
+    size_t max_wi_dims[3];
+    OCLERROR_RET(clGetDeviceInfo(device, CL_DEVICE_MAX_WORK_ITEM_SIZES,
+                                 sizeof(max_wi_dims), max_wi_dims, NULL),
+                 error, red);
+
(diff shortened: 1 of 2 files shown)


bashbaug (commented):
Interesting.  Are you able to disclose which device needs this change?

Regardless, could you please make the same change to the C++ version of this sample also, in main.cpp?  I think it would have the same problem.  Thanks!

xinjin01:
> Interesting. Are you able to disclose which device needs this change?
> 
> Regardless, could you please make the same change to the C++ version of this sample also, in main.cpp? I think it would have the same problem. Thanks!

Thanks for the review! and sorry for the delay:

This issue was observed on a device where the requested WGS exceeded the per-dimension limit (CL_DEVICE_MAX_WORK_ITEM_SIZES[0]). To improve portability across all devices, the fix clamps the WGS to the valid maximum, independent of the specific device.

I’ve also applied the same change to the C++ version (reduce.cpp), as suggested. This is included in the second commit of this PR.

bashbaug (approved):
Thanks, sorry for taking so long!  LGTM.

PR #145, opened 2025-08-27
fixes for macos CI

### LINKED ISSUE #144: Problem with macOS CI Testing
I’m not sure when it started, but there seems to be an issue with the OpenCL-SDK CI jobs on macOS, where the “Test” job is failing sporadically.  The failure looks like:
https://github.com/KhronosGroup/OpenCL-SDK/actions/runs/16787565372/job/47541877067

```
ld: library 'm' not found 
error: linker command failed with exit code 1 (use -v to see invocation)
```

If I rerun the jobs enough they seem to eventually pass, but this isn’t a viable long-term solution.  This is hurting our (already slow) ability to get SDK changes merged.

Has anyone seen an issue like this before?
If not, should we disable the macOS “Test” step for now?  We would still build on macOS, we just wouldn’t test, similar to the Android builds.

Note that macOS is a bit unique in that it is the only OS that tests using PoCL.

I also see that the macOS runners were updated to macOS 15 earlier this week - it's possible we need some updates to be compatible with macOS 15.

bashbaug:
fixes #144 

This PR includes two changes to fix macos CI:
1. Set the `SDKROOT` environment variable so PoCL can find system libraries, see discussion on the issue.
2. Disables the `callbackcpp` sample from testing, since it is producing segfaults for release builds only.  I'm not sure if this is a PoCL issue or a sample issue, but it will need additional debug.

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -771,6 +771,7 @@ jobs:
         # externalmemory is not supported on macos.
         # callbackcpp is unexpectedly failing (segfault) with PoCL - needs more debug.
         # multidevice is unexpectedly failing (clCreateSubBuffer error) with PoCL - needs more debug.
+        # see: https://github.com/KhronosGroup/OpenCL-SDK/issues/146
         EXCLUDE_REGEX="(externalmemory|callbackcpp|multidevice).*"
         ctest -C Debug --output-on-failure --no-tests=error --parallel `sysctl -n hw.logicalcpu` --exclude-regex "$EXCLUDE_REGEX"
         ctest -C Release --output-on-failure --no-tests=error --parallel `sysctl -n hw.logicalcpu` --exclude-regex "$EXCLUDE_REGEX"
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -695,17 +695,13 @@ jobs:
       run: |
         # Install Ninja only if it's the selected generator and it's not available.
         if [[ "${{matrix.GEN}}" == "Ninja Multi-Config" && ! `which ninja` ]]; then brew install ninja; fi &&
-        # We need to provide an OpenCL driver for Intel CPU on mac
-        brew install pocl
-        POCL_INSTALL_PATH=$(brew --prefix pocl)
-        echo "OCL_ICD_VENDORS=$POCL_INSTALL_PATH/etc/OpenCL/vendors" >> $GITHUB_ENV
         cmake --version
 
-    - name: Install samples dependencies
+    - name: Install Sample Dependencies (Homebrew)
       run: |
         brew install vulkan-loader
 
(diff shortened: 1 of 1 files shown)


bashbaug:
Sigh... of course now there's a completely new error I didn't see in any of my testing.

I'll need to look at this again.

OBITOONDEADO1MZ (approved):

tneval:
@bashbaug how is it that this wasn't caught in
 https://github.com/bashbaug/OpenCL-SDK/actions/runs/17270551550/job/49013861064 ?

bashbaug:
> how is it that this wasn't caught

Yeah, it's frustrating to me and it doesn't make any sense 😄 

I was hacking a bit on the CI files in my branch while testing, but it should have only been to enable testing on my branch and to disable the configs I wasn't modifying.  I don't think I did anything that would have caused this failure to go away, or anything in this PR that would cause it to appear.

Note that the "simple readme build" passed on macos for this PR, so it has something to do with the more complicated testing.  Hopefully I'll get more time to look at it in the next couple of days.

tneval:
> > how is it that this wasn't caught
> 
> Yeah, it's frustrating to me and it doesn't make any sense 😄
> 
> I was hacking a bit on the CI files in my branch while testing, but it should have only been to enable testing on my branch and to disable the configs I wasn't modifying. I don't think I did anything that would have caused this failure to go away, or anything in this PR that would cause it to appear.
> 
> Note that the "simple readme build" passed on macos for this PR, so it has something to do with the more complicated testing. Hopefully I'll get more time to look at it in the next couple of days.

Heh, I see. Was just thinking that it's possible that the env var 'hack' doesn't work with all MacOS/Xcode versions.

bashbaug:
OK, it looks like it's all building and passing if I move the PoCL set up to its own step.  This could just be coincidence, but I'll take it!  I'll mark this PR as "ready for review" again.

.github/workflows/presubmit.yml
```diff
@@ -758,10 +754,24 @@ jobs:
...
+        SDKROOT=$(xcrun --show-sdk-path)
+        echo "SDKROOT=$SDKROOT" >> $GITHUB_ENV
+
     - name: Test
       working-directory: ${{runner.workspace}}/OpenCL-SDK/build
       run: |
-        EXCLUDE_REGEX="(multidevice|externalmemory).*"
+        # externalmemory is not supported on macos.
+        # callbackcpp is unexpectedly failing (segfault) with PoCL - needs more debug.
+        # multidevice is unexpectedly failing (clCreateSubBuffer error) with PoCL - needs more debug.
```
bashbaug, line 773:
```suggestion
        # multidevice is unexpectedly failing (clCreateSubBuffer error) with PoCL - needs more debug.
        # see: https://github.com/KhronosGroup/OpenCL-SDK/issues/146
```

kpet (approved):

Issue #158, opened 2026-08-18
blur samples are failing in Linux CI


bashbaug:
The `blur` and `blurcpp` samples are failing in Linux CI.  I am not sure why.  They are passing for me locally, on all of the OpenCL implementations I've tried.  So, it is possible that this is a sample issue, though it is more likely that it is an issue with the OpenCL implementation we are testing with in CI.

PR #159, opened 2026-08-19
fix CI

bashbaug:
Fixes various issues preventing CI from passing on the latest GitHub runners:

* Updates to Visual Studio 2026 instead of Visual Studio 2022.
* Removes the explicit CMake install step and uses the (newer) default CMake and ninja instead.
* Updates the version for several third-party dependencies (freetype, GLEW, glm). The older versions did not work with a newer CMake.
* Removes the output directory overrides for several third-party dependencies that no longer work with the updated versions. Removes a few more output directory overrides that are not required, for consistency.
* Only build the static GLEW library. The shared version was not building in CI, and the static version is the only version that is needed,
* Disable the blur tests from Linux CI, since they are failing unexpectedly (issue filed: https://github.com/KhronosGroup/OpenCL-SDK/issues/158).

Note, I also enabled CI on pushes to branches other than `main`.  I can revert this if desired, but it will mean that CI will only run when a PR is created.


rjodinchr (approved):

bashbaug:
Merging offline, as discussed in the September 1st teleconference.

